Issue #46, opened 2023-03-14
Align with latest decisions (early March 2023)


ftomassetti:
"Inspired" by https://github.com/LIonWeb-org/lioncore-typescript/issues/48

- [x] https://github.com/LIonWeb-org/organization/issues/93
- [x] https://github.com/LIonWeb-org/organization/issues/92 (waiting for decision)
- [x] https://github.com/LIonWeb-org/organization/issues/91 (PR #56)
- [x] https://github.com/LIonWeb-org/organization/issues/90
- [x] https://github.com/LIonWeb-org/organization/issues/89
- [x] https://github.com/LIonWeb-org/organization/issues/86 (not yet confirmed)

ftomassetti:
I should change the type of Metamodel.version from string to integer

dslmeinte:
Didn't we agree for `Metamodel.version` to be a `string`?

enikao:
It might even be both:
* `integer` in M3
* `string` in all places in serialization format

(just showing options here, not an expression of opinion/preference)

PR #47, opened 2023-03-20
Introducing M3 changes for recent changes

ftomassetti:
As a first step towards implementing #46 we introduced some changes in M3. In this intermediate state we have a few tests failing. They should eventually pass as we include all the changes which are stored in the larger branch `issue46`

### DIFF
--- a/core/src/main/java/org/lionweb/lioncore/java/metamodel/Metamodel.java
+++ b/core/src/main/java/org/lionweb/lioncore/java/metamodel/Metamodel.java
@@ -35,7 +35,7 @@ public class Metamodel extends M3Node<Metamodel> implements NamespaceProvider, H
         return this;
     }
 
-    public Metamodel setVersion(@Nullable Integer version) {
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/lionweb/lioncore/java/metamodel/Metamodel.java
```diff
@@ -35,6 +35,17 @@ public Metamodel setName(String name) {
         return this;
     }
 
+    public Metamodel setVersion(@Nullable Integer version) {
```
dslmeinte, line 38: Didn't we decide to make a Metamodel's version a String?

dslmeinte (approved):

core/src/main/java/org/lionweb/lioncore/java/metamodel/Metamodel.java
ftomassetti: You are right!

PR #48, opened 2023-03-21
Introducing LowLevelSerialization

ftomassetti:
With this PR we introduce the LowLevelSerialization. It is intended to load semantically invalid/incoherent chunks and it is an intermediate rappresentation between the "proper" models and metamodels and the JSON representation.

This PR includes an update of all examples.

Some tests are still failing and they will be corrected as part of the remaining PR to finish up #46


ftomassetti:
I am adding both @enikao and @dslmeinte as reviewers: I think one review would be sufficient so I think once we got one in it is not strictly needed to have also the other one

dslmeinte:
Big PR! It looks good. The tests (in this repo) seem to be failing currently, though.

ftomassetti:
Yes, it is a big PR, but I obtained by splitting the original PR in three parts :D This was the second one of the three and third one is coming next. The tests are failing because I need to open a separate PR for the changes in the high-level serialization class. With the next PR all tests should be passing

PR #51, opened 2023-03-25
taught gradle to use utf-8 for source file encoding

enikao:
also removed extended codepoint from sources


ftomassetti:
Thank you!

PR #55, opened 2023-04-07
Making keys in LIonCore shorter

### LINKED ISSUE #54: Use shorter keys in LIonCore M3
Citing @dslmeinte 

> I saw you set key = id in the LIonCore M3, but I'd say that it's possible to just use the (simple/short) name as key (as long as it's unique), because the meta-pointers provide enough information. The shorter the key, the more readable the serialization, and for the M3 we can check whether it's unique upfront.

This issue is for remembering to apply this change in the Java implementation

ftomassetti:
Fix #54

### DIFF
--- a/core/src/test/resources/serialization/lioncore.json
+++ b/core/src/test/resources/serialization/lioncore.json
@@ -7,14 +7,14 @@
       "concept": {
         "metamodel": "LIonCore_M3",
         "version": "1",
-        "key": "LIonCore_M3_Metamodel"
+        "key": "Metamodel"
       },
       "properties": [
         {
           "property": {
             "metamodel": "LIonCore_M3",
             "version": "1",
-            "key": "LIonCore_M3_Metamodel_name"
+            "key": "Metamodel_name"
           },
           "value": "LIonCore.M3"
         },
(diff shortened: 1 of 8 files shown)


core/src/test/resources/serialization/lioncore.json
```diff
@@ -80,48 +80,48 @@
...
+            "key": "Concept_abstract"
           },
           "value": null
         },
         {
           "property": {
             "metamodel": "LIonCore_M3",
             "version": "1",
-            "key": "LIonCore_M3_NamespacedEntity_simpleName"
+            "key": "NamespacedEntity_simpleName"
```
dslmeinte, line 98: Hadn't we renamed this to `"name"`?

Also: these keys are still pretty qualified. Apart from the ones with obvious clashes, I'd go for the shortest, human-readable key possible. But I have to adapt the TS implementation still to see whether that's sensible.

ftomassetti: I renamed simpleName in name and qualified only the keys that we needed to qualify. This required a lot of fixing existing metamodels :)

ftomassetti: Ok, something went wrong and I lost the changes...

ftomassetti:
All comments addressed, so I am merging

PR #56, opened 2023-04-07
Verifying the format of IDs and Keys

ftomassetti:


core/src/main/java/org/lionweb/lioncore/java/utils/ModelValidator.java
```diff
@@ -0,0 +1,45 @@
...
+     *
+     * Ids can only contain these symbols:
+     *
+     * - lowercase latin characters: a..z
+     * - uppercase latin characters: A..Z
+     * - arabic numerals: 0..9
+     * - underscore: _
+     * - hyphen: -
+     */
+    public static boolean isValidID(String id) {
```
dslmeinte, line 24: Can't this be a static method on some other class? Then the `ModelValidator` class is essentially superfluous.

core/src/main/java/org/lionweb/lioncore/java/utils/ValidationResult.java
```diff
@@ -0,0 +1,28 @@
+package org.lionweb.lioncore.java.utils;
```
dslmeinte, line 1: I think we envisioned that validation results/issues/etc. would also be a model, so we should define an M2 for this. That also means we can just reference the offending node through `id`.

core/src/main/java/org/lionweb/lioncore/java/utils/ModelValidator.java
```diff
@@ -0,0 +1,45 @@
...
+        if (id.length() == 0) {
+            return false;
+        }
+        for (char c : id.toCharArray()) {
```
dslmeinte, line 28: I think I'd prefer a proper regex for this, as that's generally more portable.

core/src/main/java/org/lionweb/lioncore/java/utils/ModelValidator.java
ftomassetti: Ok, I will move it to MetamodelValidator and recreate ModelValidator later on

core/src/main/java/org/lionweb/lioncore/java/utils/ValidationResult.java
ftomassetti: Right, but I think we did not yet define rules for that, so I would refactor this code once we get the validation m2

core/src/main/java/org/lionweb/lioncore/java/utils/ModelValidator.java
ftomassetti: Changing it to a regex

PR #59, opened 2023-04-27
Changes necessary to make possible to unserialize ASTs in Kolasu

ftomassetti:
The changes permits to unserialize nodes directly by using the values of properties and children and construction time.
To do so we want to start instantiating the nodes starting from the leaves and we want to provide the values of properties and children to the custom unserializers

### DIFF
--- a/core/src/main/java/org/lionweb/lioncore/java/api/CompositeNodeResolver.java
+++ b/core/src/main/java/org/lionweb/lioncore/java/api/CompositeNodeResolver.java
@@ -36,6 +36,6 @@ public class CompositeNodeResolver implements NodeResolver {
 
   @Override
   public String toString() {
-    return "CompositeNodeResolver(" + nodeResolvers +")";
+    return "CompositeNodeResolver(" + nodeResolvers + ")";
   }
 }
--- a/core/src/main/java/org/lionweb/lioncore/java/api/CompositeNodeResolver.java
+++ b/core/src/main/java/org/lionweb/lioncore/java/api/CompositeNodeResolver.java
@@ -33,4 +33,9 @@ public class CompositeNodeResolver implements NodeResolver {
     }
     return null;
   }
+
+  @Override
+  public String toString() {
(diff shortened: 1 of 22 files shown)


enikao:
Why did you chose the bottom-up processing? I presume to be sure all referenced ids are present.

The alternative would be a two-phase approach: first read all nodes, then resolve them.

How do you handle invalid models (multiple roots, containment loops, dead references)?

ftomassetti:
Yes, for resolving references we use the two-phases approach, but in this case the ordering is necessary so that children are already unserialized when instantiating the parent. This is useful because unserializer could potentially want to pass the children (already unserialized) to the constructor of the parent. This is optional, and if this is not done and an empty constructor is available it will be used and children will be added after instantiation. However this approach permits to suppor the two cases and having the non-empty constructor is very convenient in certain cases.

Containment loops are not supported, while references loop would be ok. Dead references (if they mean references to stuff we cannot resolve) would cause an issue for the "high level serialization", while they would work for the "low level serialization"

enikao:
> Yes, for resolving references we use the two-phases approach, but in this case the ordering is necessary so that children are already unserialized when instantiating the parent. This is useful because unserializer could potentially want to pass the children (already unserialized) to the constructor of the parent. This is optional, and if this is not done and an empty constructor is available it will be used and children will be added after instantiation. However this approach permits to suppor the two cases and having the non-empty constructor is very convenient in certain cases.

Makes sense.

> Containment loops are not supported, while references loop would be ok. Dead references (if they mean references to stuff we cannot resolve) would cause an issue for the "high level serialization", while they would work for the "low level serialization"

My _actual_ question on this was missing: Are we sure this implementation handles these cases as expected?

ftomassetti:
> My _actual_ question on this was missing: Are we sure this implementation handles these cases as expected?

Good point. I have added 8 new tests to verify more corner cases and that leads to fix some issues.

I checked for dead references, references loops, containment loops, using null IDs for either parents or children, unserializing multiple roots

PR #62, opened 2023-05-05
Better error message for unserializing tree without root

ftomassetti:


ftomassetti:
Merging right away, as we are working on the demo and we want to move quickly

PR #67, opened 2023-05-06
Adding enumeration literal key and using it for serialization

### LINKED ISSUE #53: Why does EnumerationLiteral not have getKey()?
All other M3 members have such a method.

ftomassetti:
Should fix #53

### DIFF
--- a/core/src/main/java/io/lionweb/lioncore/java/serialization/PrimitiveValuesSerialization.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/serialization/PrimitiveValuesSerialization.java
@@ -14,7 +14,8 @@ import java.util.*;
  * of the primitive value.
  */
 public class PrimitiveValuesSerialization {
-
+  // We use the ID, and not the key, to classify the enumerations internally within PrimitiveValuesSerialization
+  // because that is unique
   private Map<String, Enumeration> enumerationsByID = new HashMap<>();
 
   public void registerMetamodel(Metamodel metamodel) {
@@ -91,6 +92,9 @@ public class PrimitiveValuesSerialization {
       if (serializedValue == null) {
         return null;
       }
+      // In this case, where we are dealing with primitive values, we want to use the literal _key_ (and not the ID)
+      // This is at least the default behavior, but the user can register specialized primitiveUnserializers,
+      // if a different behavior is needed
       Optional<EnumerationLiteral> enumerationLiteral =
           enumerationsByID.get(primitiveTypeID).getLiterals().stream()
               .filter(l -> Objects.equals(l.getKey(), serializedValue))
@@ -114,6 +118,9 @@ public class PrimitiveValuesSerialization {
       if (value == null) {
         return null;
       }
+      // In this case, where we are dealing with primitive values, we want to use the literal _key_ (and not the ID)
+      // This is at least the default behavior, but the user can register specialized primitiveSerializers,
+      // if a different behavior is needed
       return ((EnumerationLiteral) value).getKey();
     } else {
       throw new IllegalArgumentException(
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/lionweb/lioncore/java/serialization/PrimitiveValuesSerialization.java
```diff
@@ -3,15 +3,26 @@
...
  * of the primitive value.
  */
 public class PrimitiveValuesSerialization {
+
+  private Map<String, Enumeration> enumerationsByID = new HashMap<>();
+
+  public void registerMetamodel(Metamodel metamodel) {
+    metamodel.getElements().stream()
+        .filter(e -> e instanceof Enumeration)
+        .forEach(e -> enumerationsByID.put(e.getID(), (Enumeration) e));
```
enikao, line 23: enumeration keys are only required to be unique within the language, so we could have a key clash here.

core/src/main/java/io/lionweb/lioncore/java/serialization/PrimitiveValuesSerialization.java
```diff
@@ -3,15 +3,26 @@
...
+import io.lionweb.lioncore.java.metamodel.Metamodel;
+import java.util.*;
 
 /**
  * This class is responsible for serialization and unserializing primitive values, based on the type
  * of the primitive value.
  */
 public class PrimitiveValuesSerialization {
+
+  private Map<String, Enumeration> enumerationsByID = new HashMap<>();
```
enikao, line 18: `enumerationsByKey`?

core/src/main/java/io/lionweb/lioncore/java/serialization/PrimitiveValuesSerialization.java
```diff
@@ -76,6 +87,19 @@ public void registerLionBuiltinsPrimitiveSerializersAndUnserializers() {
   public Object unserialize(String primitiveTypeID, String serializedValue) {
```
enikao, line 87: `primitiveTypeKey`?

core/src/main/java/io/lionweb/lioncore/java/serialization/PrimitiveValuesSerialization.java
```diff
@@ -95,4 +124,8 @@ public String serialize(String primitiveTypeID, Object value) {
               + ")");
     }
   }
+
+  private boolean isEnum(String primitiveTypeID) {
```
enikao, line 128: `primitiveTypeKey`?

core/src/main/java/io/lionweb/lioncore/java/serialization/PrimitiveValuesSerialization.java
ftomassetti: Yes, but here in enumerationsByID we want to use the ID and not the key, so that we are able to retrieve the enumeration definitions easily. We then use the key for the value. EnumerationsByID is only an internal structure used by PrimitiveValuesSerialization (I am adding a comment about that)

core/src/main/java/io/lionweb/lioncore/java/serialization/PrimitiveValuesSerialization.java
ftomassetti: No, I think we want to use the ID for quickly and uniquely identify the enumerations (we still use the key to unserialize values)

core/src/main/java/io/lionweb/lioncore/java/serialization/PrimitiveValuesSerialization.java
ftomassetti: To clarify: yes, we use the enumeration key when figuring out which enumeration we are dealing with. So at some point we get an Enumeration instance (i.e., we have access to the name, id, key, etc.).

Then in `serializePropertyValue` we call the `primitiveValuesSerialization` in this way:
```
primitiveValuesSerialization.serialize(dataType.getID(), value);
```

There we pass the primitive type ID, because it is unique. It is not the value we use to figure out the primitive type, but just as unique identifier to register and then retrieve the primitive serializers and unserializers

enikao: You're right.

enikao (approved):

PR #79, opened 2023-07-17
Added M3Node.toString

enikao:
Also applied some refactorings proposed by IntelliJ


ftomassetti:
Looks good, thanks. Please let me know if you need me to do a release

ftomassetti (approved):
+1

ftomassetti:
Yes, I think would be nice to unify them. I think that both the ide and the concept name are useful information to have

ftomassetti:
Let's start with merging this and then I will look into the toString methods. Thank you @enikao !

ftomassetti:
As a follow-up I created https://github.com/LIonWeb-org/lioncore-java/pull/80

PR #80, opened 2023-07-20
Revising the toString methods for descendants of M3Node

ftomassetti:
They now all produce the same initial part and some add additional details

### DIFF
--- a/core/src/main/java/io/lionweb/lioncore/java/language/Containment.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/language/Containment.java
@@ -100,7 +100,7 @@ public class Containment extends Link<Containment> {
 
   @Override
   public String toString() {
-    return "Containment{" + "name=" + getName() + ", " + "type=" + getType() + '}';
+    return super.toString() + "{" + "name=" + getName() + ", " + "type=" + getType() + '}';
   }
 
   @Override
--- a/core/src/main/java/io/lionweb/lioncore/java/language/Containment.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/language/Containment.java
@@ -99,11 +99,6 @@ public class Containment extends Link<Containment> {
   }
 
   @Override
-  public String toString() {
-    return super.toString() + "{" + "name=" + getName() + ", " + "type=" + getType() + '}';
-  }
-
-  @Override
   public Concept getConcept() {
     return LionCore.getContainment();
   }
--- a/core/src/main/java/io/lionweb/lioncore/java/language/PrimitiveType.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/language/PrimitiveType.java
@@ -37,7 +37,7 @@ public class PrimitiveType extends DataType<PrimitiveType> {
 
   @Override
   public String toString() {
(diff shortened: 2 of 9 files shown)


core/src/main/java/io/lionweb/lioncore/java/language/Containment.java
```diff
@@ -100,7 +100,7 @@ public void setSpecialized(@Nullable Containment specialized) {
 
   @Override
   public String toString() {
-    return "Containment{" + "name=" + getName() + ", " + "type=" + getType() + '}';
+    return super.toString() + "{" + "name=" + getName() + ", " + "type=" + getType() + '}';
```
enikao, line 103: This is the same implementation as for `Reference`, so we can move it to `Link`

core/src/main/java/io/lionweb/lioncore/java/language/PrimitiveType.java
```diff
@@ -37,7 +37,7 @@ public PrimitiveType(@Nullable Language language, @Nullable String name, String
 
   @Override
   public String toString() {
-    return "PrimitiveType(" + getName() + ")";
+    return super.toString() + "{" + "name=" + getName() + '}';
```
enikao, line 40: This is the only `LanguageEntity` that has a special `toString()`. Why?

core/src/main/java/io/lionweb/lioncore/java/language/Containment.java
enikao: Why don't construct a _qualifiedName_ like for `LanguageEntity` (for symmetry)?

core/src/main/java/io/lionweb/lioncore/java/language/Property.java
```diff
@@ -73,7 +73,7 @@ public Property setType(@Nullable DataType type) {
 
   @Override
   public String toString() {
-    return "Property{" + "name=" + getName() + ", " + "type=" + getType() + '}';
+    return super.toString() + "{" + "name=" + getName() + ", " + "type=" + getType() + '}';
```
enikao, line 76: Why don't construct a _qualifiedName_ like for `LanguageEntity` (for symmetry)?

core/src/main/java/io/lionweb/lioncore/java/language/PrimitiveType.java
ftomassetti: Right, let's remove it

core/src/main/java/io/lionweb/lioncore/java/language/Containment.java
ftomassetti: Moving to Link, and using qualifiedName with the same logic used in LanguageEntity. I will also use qualifiedName in Property.toString

core/src/main/java/io/lionweb/lioncore/java/language/Property.java
ftomassetti: Done

ftomassetti:
Thank you @enikao ! I applied all of your suggestions and I will now merge this

PR #84, opened 2023-09-08
Annotations

ftomassetti:
With this PR:
* We introduce annotations in the language (i.e., the meta-metamodel)
* We introduce annotations in the model API. To do this we create a common ancestor for annotations and nodes as annotations can target both. We call this ancestor "Element" for lack of a better name
* We start revising the serialization support not to explode if "annotations" is encountered
* Wrote tests regarding the model API
* Added support for actually exporting annotations in the serialization format 

What we still need to do:
* Write tests add support in the EMF importer/exporter (https://github.com/LIonWeb-org/lioncore-java/issues/85)


enikao:
> * We introduce annotations in the model API. To do this we create a common ancestor for annotations and nodes as annotations can target both. We call this ancestor "Element" for lack of a better name

Isn't the common ancestor `Classifier`? (See https://lionweb-org.github.io/organization/metametamodel/metametamodel.html#_overview)

ftomassetti:
>> We introduce annotations in the model API. To do this we create a common ancestor for annotations and nodes as annotations can target both. We call this ancestor "Element" for lack of a better name

>Isn't the common ancestor Classifier? (See https://lionweb-org.github.io/organization/metametamodel/metametamodel.html#_overview)

It would be the equivalent on Classifier on the model level. I.e., it would have the thing that represent an instance of a Classifier, in the same way that Node is an instance on Concept and AnnotationInstance is an instance of Annotation

ftomassetti:
I renamed `Element` into `ClassifierInstance`, because @enikao is right, is connected to `Classifier`

ftomassetti:
Discussion in https://github.com/LIonWeb-org/organization/issues/164 could affect this

ftomassetti:
This is a very big PR because a lot of code had to be changed to deal with Classifiers (and not just Concepts) and with Classifier Instances (which are either Nodes or Annotation Instances, while before we dealt only with Node).

I do not expect anyone to have the time to read this whole PR, but perhaps @enikao and/or @dslmeinte have the time to take a look at some part and provide some feedback.

Otherwise I can merge and then react to issues and inconsistencies with the TS implementation as they emerge :)

dslmeinte:
It's indeed a big PR. What I'm sort of missing is a persisted serialization chunk with some actual annotation instances. I only see the key-value-pair `"annotations": []` having been added everywhere: that might be something to add - or likely explicitly deserialize a language + model in it that's built up in some of the tests.

dslmeinte (approved):

core/src/main/java/io/lionweb/lioncore/java/serialization/ClassifierResolver.java
```diff
@@ -50,4 +64,8 @@ public ConceptResolver registerLanguage(Language language) {
   private void registerConcept(Concept concept) {
     registeredConcepts.put(MetaPointer.from(concept), concept);
   }
+
+  private void registeredAnnotation(Annotation annotation) {
```
enikao, line 67: typo: `registerAnnotation`

core/src/main/java/io/lionweb/lioncore/java/serialization/data/SerializedClassifierInstance.java
```diff
@@ -160,4 +164,12 @@ public String toString() {
     public void setAnnotations(List<String> annotationIDs) {
       this.annotations = annotationIDs;
     }
+
+    public String getAnnotated() {
```
enikao: Why do we have this method? This _must_ be the same as `getParent()`, according to https://lionweb-org.github.io/organization/serialization/serialization.html#parent

core/src/main/java/io/lionweb/lioncore/java/serialization/data/SerializedClassifierInstance.java
```diff
@@ -11,6 +11,7 @@ public class SerializedClassifierInstance {
   private String id;
```
enikao: What's the difference to a `SerializedNode`?

core/src/main/java/io/lionweb/lioncore/java/model/AnnotationInstance.java
```diff
@@ -1,13 +1,18 @@
...
- * links.
+ * While an AnnotationInstance implements HasFeatureValues, it is forbidden to hold any children, as
+ * the Annotation should not have any containment link.
  */
-@Experimental
-public interface AnnotationInstance extends HasFeatureValues {
+public interface AnnotationInstance extends ClassifierInstance<Annotation> {
   Annotation getAnnotationDefinition();
+
+  ClassifierInstance getAnnotated();
```
enikao, line 13: same as `getParent()`

core/src/main/java/io/lionweb/lioncore/java/serialization/ClassifierResolver.java
ftomassetti: fixed, thank you

ftomassetti:
TODO: let annotations have containments

core/src/main/java/io/lionweb/lioncore/java/serialization/data/SerializedClassifierInstance.java
ftomassetti: This came as a surprise to me. When I think about the model I imagine a tree (or a forest) which is composed by nodes, each one with their own position within the tree. Each one (besides roots) having a parent. Annotation instances are instances something else: they are not node, they are not part of the tree. They are attached to nodes. Some consumers are interested in the tree (so just the nodes), some others are interested in the tree and the attached annotation instances.

So it seems strange to me that an annotation instance has a parent, as it is not part of the tree. From may point of view it has a pointer to the node that it is attached to, which it is not its parent, as the annotation instance is not a children of the node (it is not part of any containment).

core/src/main/java/io/lionweb/lioncore/java/serialization/data/SerializedClassifierInstance.java
ftomassetti: In my understanding we have annotation instances and nodes, which are similar but not the same. So the common ancestor (grouping the similarities) is called Classifier Instance. Also when serializing them they are very similar but not the same. For example, an annotation instance has no concept, it has an annotation instead. It has no parent but an annotated node instead (see previous comment). For this reason also on the serialization level I find useful to distinguish them.

core/src/main/java/io/lionweb/lioncore/java/serialization/data/SerializedClassifierInstance.java
enikao: The rationale was: Every node _must_ be contained in a `Partition`. We can always find the partition by moving up to the root of the containment hierarchy.
If not the annotated node, who else is the parent of an annotation instance? Which partition does it belong to?

If we wanted _completely independent_ annotations (i.e. stored somewhere outside the annotated node), they would be provided by a processor as a derived model. This processor might have its own internal storage, or ask the repository for a "special" partition that contains all the annotations.

ftomassetti: I think that the partition containing the annotation is the same partition which is containing the annotated node, on that we agree. My problem is with the term "parent" to indicate what it seems to me are two different relationship.

I think that the children of a node do not contain the annotations attached to that node. Is that right?

If it is right, then it is surprising that the annotation can have as a parent the annotated node. It means that parent is not the inverse relationship of children.

enikao: That's what we decided in https://github.com/LIonWeb-org/organization/issues/150

enikao: I hope this was _our_ understanding, not only _my_ understanding -- shall we raise the issue again in the next meeting?

ftomassetti:
Merging to avoid conflicts with other work (lke renaming LIonWeb into LionWeb).

Some changes may be done later, to ensure adherence with the "standard"

PR #92, opened 2023-10-10
rename concept classifier and change serializationFormatVersion to 2023.1

enikao:


enikao:
So far only contains the changes to JSON files, I might add the code changes later.

enikao:
included in already merged #94

PR #94, opened 2023-10-11
code optimization

enikao:
depends on #92 

Mostly followed IntelliJ's proposals. Tried to put every category of fixes into a separate commit, so we can accept or rollback separately.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -146,10 +146,10 @@ java {
     targetCompatibility = JavaVersion.toVersion(jvmVersion)
 }
 
-if (isReleaseVersion) {
-    tasks.withType(Sign::class) {
-    }
+tasks.withType(Sign::class) {
+    onlyIf("isReleaseVersion is set", { isReleaseVersion } )
 }
+
 signing {
     sign(publishing.publications["lioncore_java_core"])
 }
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -147,7 +147,7 @@ java {
 }
 
 tasks.withType(Sign::class) {
-    onlyIf("isReleaseVersion is set", { isReleaseVersion } )
+    onlyIf("isReleaseVersion is set") { isReleaseVersion }
 }
 
 signing {
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -99,12 +99,12 @@ publishing {
                 description.set("Java APIs for the LionWeb system")
                 version = project.version as String
                 packaging = "jar"
-                url.set("https://github.com/LionWeb-org/lioncore-java")
+                url.set("https://github.com/LionWeb-io/lioncore-java")
 
                 scm {
-                    connection.set("scm:git:https://github.com/LionWeb-org/lioncore-java.git")
-                    developerConnection.set("scm:git:git@github.com:LionWeb-org/lioncore-java.git")
-                    url.set("https://github.com/LionWeb-org/lioncore-java.git")
+                    connection.set("scm:git:https://github.com/LionWeb-io/lioncore-java.git")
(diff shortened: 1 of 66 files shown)


core/build.gradle.kts
```diff
@@ -146,10 +146,10 @@ java {
     targetCompatibility = JavaVersion.toVersion(jvmVersion)
 }
 
-if (isReleaseVersion) {
-    tasks.withType(Sign::class) {
-    }
+tasks.withType(Sign::class) {
+    onlyIf("isReleaseVersion is set", { isReleaseVersion } )
```
ftomassetti, line 150: I think it may be idiomatic in Kotlin to put a lambda appearing as last parameter outside the parenthesis
```suggestion
    onlyIf("isReleaseVersion is set") { isReleaseVersion }
```

core/src/main/java/io/lionweb/lioncore/java/language/Classifier.java
```diff
@@ -50,6 +50,7 @@ public Classifier(@Nullable Language language, @Nullable String name) {
 
   public @Nonnull List<Classifier<?>> allAncestors() {
     List<Classifier<?>> ancestors = new ArrayList<>();
+    // TODO this implementation seems to be wrong
```
ftomassetti, line 53: This is indeed very wrong. But probably we expect this to be overriden. Maybe this should just be abstract

core/src/main/java/io/lionweb/lioncore/java/self/LionCore.java
```diff
@@ -77,7 +77,7 @@ public static Language getInstance() {
       INSTANCE = new Language("LionCore.M3");
       INSTANCE.setID("-id-LionCore-M3");
       INSTANCE.setKey("LionCore-M3");
-      INSTANCE.setVersion("1");
+      INSTANCE.setVersion("2023.1");
```
ftomassetti, line 80: Should we refer to the constant defined in JsonSerialization or viceversa defining it here and refer to it from JsonSerialization?

core/src/test/java/io/lionweb/lioncore/java/serialization/SerializedJsonComparisonUtils.java
```diff
@@ -84,36 +84,46 @@ private static void assertEquivalentNodes(
...
-            "Properties of " + context);
-      } else if (key.equals("annotations")) {
-        assertEquivalentUnorderedArrays(
-            expected.getAsJsonArray("annotations"),
-            actual.getAsJsonArray("annotations"),
-            "Annotations of " + context);
-      } else {
-        throw new AssertionError("(" + context + ") unexpected top-level key found: " + key);
+      switch (key) {
+        case "parent":
```
ftomassetti, line 88: Is this supported by Java 8? I did not expect that

ftomassetti: By this I mean, switch on string constants

ftomassetti (approved):
Thank you, it looks great to me

core/src/test/java/io/lionweb/lioncore/java/serialization/SerializedJsonComparisonUtils.java
enikao: Yes, all checks passed for this PR.
Even Java 7 supported it: https://docs.oracle.com/javase/8/docs/technotes/guides/language/strings-switch.html

core/src/main/java/io/lionweb/lioncore/java/language/Classifier.java
enikao: Implemented a fixed generic version

PR #99, opened 2023-10-20
When populating EClasses from Concepts, if there are implementing concept an exception occurs

AtanasM98:
Adjusted code to map implemented classes to EClasses and add them to the superTypes of the EClass in question.


ftomassetti:
Thank you forr the PR

ftomassetti:
P.S. When submitting a PR can be useful to run `./gradlew spotlessApply` to run the formatter, otherwise the build could fail

PR #110, opened 2023-10-30
trial to get signing to work on windows

enikao:


dslmeinte:
Doesn't look like it did... :(

enikao:
> Doesn't look like it did... :(

First need to merge, can only release from master

Issue #111, opened 2023-11-06
Single Containment reference is not supported


AtanasM98:
Trying to export a model that uses a meta model that has a reference set as a single containment(Lower & Upper bound=1) throws and unsupported exception. [Example of code with this problem.](https://github.com/LionWeb-io/property-demo/tree/atanas/single-containment-bug)
[Line in EMFModelExporter where problem occurs.](https://github.com/LionWeb-io/lionweb-java/blob/f6a67bdd1a7dd54c032dedb9a95d0458cdc4ad25/emf/src/main/java/io/lionweb/lioncore/java/emf/EMFModelExporter.java#L67)
## How to reproduce bug:
In property-demo repository execute
`./gradlew xtend-generators:emf-generator:run`

ftomassetti:
How urgent is this? Do you need solved for LangDev?

ftomassetti:
I think I have a fix for this, but to test it could you share with me the \example1-exported.lm.json file referred in the linked repo?

AtanasM98:
Hello sorry. I commited the [lm file in the repo.](https://github.com/LionWeb-io/property-demo/blob/atanas/xtend-generators/properties-parser/example1-exported.lm.json)

ftomassetti:
I cannot reproduce the problem, but given we are on a deadline I will open the PR with the fix anyway

ftomassetti:
I have now a test for this, but my test is currently not passing even after my fix for a problem I cannot figure out. At some point this line is called with eClass=Property:

```
EObject eObject = eClass.getEPackage().getEFactoryInstance().create(eClass);
```

Given that Property implements INamed, this code cause the create method to be invoked with eClass=INamed:

```
  public EObject create(EClass eClass) 
  {
    if (getEPackage() != eClass.getEPackage() || eClass.isAbstract())
    {
      throw new IllegalArgumentException("The class '" + eClass.getName() + "' is not a valid classifier");
    }

    for (List<EClass> eSuperTypes = eClass.getESuperTypes(); !eSuperTypes.isEmpty(); )
    {
      EClass eSuperType = eSuperTypes.get(0);
      if (eSuperType.getInstanceClass() != null)
      {
        EObject result = eSuperType.getEPackage().getEFactoryInstance().create(eSuperType);
        ((InternalEObject)result).eSetClass(eClass);
        return result;
      }
      eSuperTypes = eSuperType.getESuperTypes();
    }

    return basicCreate(eClass);
  }
[...]

PR #113, opened 2023-11-07
Provide internal implementation of INamed

### LINKED ISSUE #98: LionCore-builtins.Node and INamed should be mapped to EClasses when instantiating an EMFModelExporter
We had to manually do this when trying to register a language file to EMF.

```
var ConceptsToEClassesMapping conceptMapper = new ConceptsToEClassesMapping();

conceptMapper.registerMapping(LionCoreBuiltins.getNode(), EcorePackage.eINSTANCE.getEObject())

val builtinsEPackage = EcoreFactory.eINSTANCE.createEPackage()
builtinsEPackage.name = "LionCore_builtins"
builtinsEPackage.nsURI = "LionCore-builtins"

val eCoreINamed = EcoreFactory.eINSTANCE.createEClass()
eCoreINamed.name = "INamed"
eCoreINamed.interface = true
builtinsEPackage.EClassifiers.add(eCoreINamed)

val eCoreINamedName = EcoreFactory.eINSTANCE.createEAttribute()
eCoreINamedName.name = "name"
eCoreINamedName.EType = EcorePackage.eINSTANCE.EString
eCoreINamed.EStructuralFeatures.add(eCoreINamedName)

conceptMapper.registerMapping(LionCoreBuiltins.getINamed(), eCoreINamed)
```

This should be preregistered since it's something every language is going to use.
You can see it in action in [my repository using EMF reflective access](https://github.com/f1re-nl/emf-gettingstarted/blob/main/nl.f1re.emf.generator.emf.reflective/src/main/java/nl/f1re/emf/generator/emf/reflective/EmfReflective.xtend).

enikao:
fixes #98


emf/src/main/java/io/lionweb/lioncore/java/emf/mapping/ConceptsToEClassesMapping.java
```diff
@@ -23,6 +19,10 @@ public class ConceptsToEClassesMapping {
   private final Map<Concept, EClass> conceptsToEClasses = new HashMap<>();
   private final Map<Interface, EClass> interfacesToEClasses = new HashMap<>();
 
+  public ConceptsToEClassesMapping() {
+    populateInternal();
```
enikao, line 23: Should we do this by default, or only on explicit call?

emf-builtins/.classpath
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ftomassetti, line 1: Do we need this file to be committed? If so, could we add a line in the README of this project about this file?

emf-builtins/.project
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ftomassetti, line 1: Is this an Eclipse specific file? Do we need this file to be committed? If so, could we add a line in the README of this project about this file?

emf-builtins/build.gradle
```diff
@@ -0,0 +1,154 @@
+import org.apache.tools.ant.taskdefs.condition.Os
```
ftomassetti, line 1: For the other projects we are using .gradle.kts files. Would that be feasible to move it to a .gradle.kts? If it takes any significant amount of work, can you just open an issue for a follow up task for doing that?

emf-builtins/build.gradle
```diff
@@ -0,0 +1,154 @@
...
+                    "${projectDir}/src/main/model"
+            ])
+        }
+    }
+}
+
+var isReleaseVersion = !(version as String).endsWith("SNAPSHOT")
+
+task javadocJar(type: Jar) {
+//    dependsOn("myJavadoc")
```
ftomassetti, line 65: Can we remove these lines?

emf-builtins/src/main/model/builtins.ecore
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ftomassetti, line 1: Wouldn't be simpler to define this EPackage programmatically?

ftomassetti: Is this necessary because we need the genmodel?

emf/src/main/java/io/lionweb/lioncore/java/emf/mapping/ConceptsToEClassesMapping.java
```diff
@@ -23,6 +19,10 @@ public class ConceptsToEClassesMapping {
   private final Map<Concept, EClass> conceptsToEClasses = new HashMap<>();
   private final Map<Interface, EClass> interfacesToEClasses = new HashMap<>();
 
+  public ConceptsToEClassesMapping() {
```
ftomassetti, line 22: Perhaps later we can add a flag to give the possibility not to register this, in case it is needed

emf/src/test/java/io/lionweb/lioncore/java/emf/EMFMetamodelExporterTest.java
```diff
@@ -136,4 +147,152 @@ public void exportInterfaceAndEnumeration() {
...
+    EReference props = (EReference) propertiesFile.getEStructuralFeature("props");
+    assertEquals("props", props.getName());
+    assertEquals(1, props.getLowerBound());
+    assertEquals(-1, props.getUpperBound());
+    assertEquals(property, props.getEType());
+
+    EReference valueRef = (EReference) property.getEStructuralFeature("value");
+    assertEquals("value", valueRef.getName());
+    assertEquals(1, valueRef.getLowerBound());
+    // TODO fix
```
ftomassetti, line 229: What is the issue here?

ftomassetti (commented):
Most of my comments are probably due to the fact that I did not use EMF for years

emf-builtins/.classpath
enikao: This is part of the Eclipse project. As we need Eclipse to sensibly edit the `.ecore` and `.genmodel` files, I'd like to have a proper Eclipse project.

emf-builtins/.project
enikao: I'll add something to readme.

emf-builtins/build.gradle
enikao: Atanas said the Eclipse Gradle plugin doesn't understand `.kts`. I didn't double-check, but assume he's correct.
And we definitely want to have the same build system.

emf-builtins/build.gradle
enikao: You mean the commented ones or this task at all? Commented definitely, the task I don't know -- I just stole your code (-:

emf/src/main/java/io/lionweb/lioncore/java/emf/mapping/ConceptsToEClassesMapping.java
enikao: The method is public, so interested parties could just call it.

emf/src/test/java/io/lionweb/lioncore/java/emf/EMFMetamodelExporterTest.java
enikao: setting the upper bound to -1 is the workaround for #111

emf/src/main/java/io/lionweb/lioncore/java/emf/mapping/ConceptsToEClassesMapping.java
ftomassetti: By default I would enable it, but offer an alternative constructor with a flag to control it

emf-builtins/.classpath
ftomassetti: Ok, makes sense

emf-builtins/build.gradle
ftomassetti: Right, then maybe we should add a note in the README about this

emf-builtins/build.gradle
ftomassetti: The commented ones :D

ftomassetti (approved):
Thank you @enikao , feel free to merge at any time

PR #114, opened 2023-11-08
Fix export of EMF models using single containments

### LINKED ISSUE #111: Single Containment reference is not supported

ftomassetti:
Fix #111


emf/src/main/java/io/lionweb/lioncore/java/emf/EMFModelExporter.java
```diff
@@ -64,7 +64,17 @@ public EObject exportTree(Node root, ReferencesPostponer referencesPostponer) {
                             .collect(Collectors.toList());
                     eObject.eSet(eReference, childrenInEmf);
                   } else {
-                    throw new UnsupportedOperationException();
+                    List<? extends Node> childrenInLW =
+                        root.getChildrenByContainmentName(eReference.getName());
+                    if (childrenInLW.size() > 1) {
+                      throw new IllegalStateException(
+                          "More than one child found, where up to one ");
```
enikao, line 71: I think this error message is hard to understand without the context.

emf/src/main/java/io/lionweb/lioncore/java/emf/mapping/ConceptsToEClassesMapping.java
```diff
@@ -136,7 +136,11 @@ public boolean knows(EClassifier eClassifier) {
     return null;
   }
 
-  public void populateInternal() {
+  public void prePopulateBuiltins() {
+    ePackagesToLanguages.put(
+        BuiltinsEMFPackageProvider.getEPackage(), LionCoreBuiltins.getInstance());
```
enikao, line 141: What does `BuiltinsEMFPackageProvider` do?

ftomassetti: I should have removed that. This was the thing I was using before getting the Builtins EMF Package from your PR. Removed

emf/src/main/java/io/lionweb/lioncore/java/emf/EMFModelExporter.java
ftomassetti: Right, I tried updating it

enikao (approved):

ftomassetti:
Thank you @enikao , I will do a new release right now (0.1.4) so you can use these fixes more easily

Issue #112, opened 2023-11-06
EmfMetaModelExporter should check the registry for existing eclasses before trying to create duplicates


AtanasM98:
[Code where problem occurs.](https://github.com/LionWeb-io/lionweb-java/blob/f6a67bdd1a7dd54c032dedb9a95d0458cdc4ad25/emf/src/main/java/io/lionweb/lioncore/java/emf/EMFMetamodelExporter.java#L41C15-L41C15)

ftomassetti:
How urgent is this? Do you need solved for LangDev?

enikao:
We have a workaround, by manually adding the appropriate mappings ([source code](https://github.com/LionWeb-io/property-demo/blob/8ec9c90512ec040eace9f518f86a5b122e496997/xtend-generators/emf-generator/src/main/java/generator/emf/EmfGenerated.xtend#L45))

Would be nicer for the demo if it worked.

PR #115, opened 2023-11-09
EMF fixes

enikao:
Include more files into emf-builtins artifact, so it can be recognized as Eclipse plugin.

Implemented workaround for #112.

Added more info to exceptions.

### DIFF
--- a/emf-builtins/META-INF/MANIFEST.MF
+++ b/emf-builtins/META-INF/MANIFEST.MF
@@ -1,0 +1,6 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: emf-builtins
+Bundle-SymbolicName: io.lionweb.java.emf.builtins;singleton:=true
+Bundle-Version: 0.1.0.qualifier
+Require-Bundle: org.eclipse.emf.ecore
--- a/emf-builtins/META-INF/MANIFEST.MF
+++ b/emf-builtins/META-INF/MANIFEST.MF
@@ -1,6 +1,0 @@
-Manifest-Version: 1.0
-Bundle-ManifestVersion: 2
-Bundle-Name: emf-builtins
-Bundle-SymbolicName: io.lionweb.java.emf.builtins;singleton:=true
-Bundle-Version: 0.1.0.qualifier
-Require-Bundle: org.eclipse.emf.ecore
--- a/emf/src/main/java/io/lionweb/lioncore/java/emf/EMFMetamodelExporter.java
+++ b/emf/src/main/java/io/lionweb/lioncore/java/emf/EMFMetamodelExporter.java
@@ -3,11 +3,10 @@ package io.lionweb.lioncore.java.emf;
 import io.lionweb.lioncore.java.emf.mapping.ConceptsToEClassesMapping;
 import io.lionweb.lioncore.java.emf.mapping.DataTypeMapping;
 import io.lionweb.lioncore.java.language.*;
+import java.util.List;
 import org.eclipse.emf.ecore.*;
 import org.eclipse.emf.ecore.resource.Resource;
 import org.eclipse.emf.ecore.resource.impl.ResourceImpl;
-
-import java.util.List;
 
 /** Export LionWeb's metamodels into EMF's metamodels. */
 public class EMFMetamodelExporter extends AbstractEMFExporter {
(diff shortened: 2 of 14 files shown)


emf-builtins/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,6 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: emf-builtins
+Bundle-SymbolicName: io.lionweb.java.emf.builtins;singleton:=true
+Bundle-Version: 0.1.0.qualifier
```
ftomassetti, line 5: Is this the version of the Maven package? If so, we should consider generating this file

ftomassetti: Looking at the build.gradle file it seems generated, so I think we should not commit it, but adding the task generating this file as a dependency of the `setup` task

emf/src/main/java/io/lionweb/lioncore/java/emf/mapping/ConceptsToEClassesMapping.java
```diff
@@ -121,6 +136,28 @@ public void registerMapping(Interface iface, EClass eClass) {
     interfacesToEClasses.put(iface, eClass);
   }
 
+  public void registerMapping(Language language, EPackage ePackage) {
+    for (LanguageEntity entity : language.getElements()) {
```
ftomassetti, line 140: Should we check if this mapping already exists and in that case just return?
Should we check for non-conflict mapping?

emf/src/test/resources/properties.ecore
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <ecore:EPackage xmi:version="2.0" xmlns:xmi="http://www.omg.org/XMI" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xmlns:ecore="http://www.eclipse.org/emf/2002/Ecore" name="io.lionweb.Properties" nsURI="https://lionweb.io/io_lionweb_Properties"
+    xmlns:ecore="http://www.eclipse.org/emf/2002/Ecore" name="io_lionweb_Properties" nsURI="https://lionweb.io/io_lionweb_Properties"
```
ftomassetti, line 3: Wouldn't be better to keep the dots? Or is there a reason to replace them?

ftomassetti:
Thank you for the ton of improvements!

ftomassetti (approved):

emf/src/main/java/io/lionweb/lioncore/java/emf/EMFMetamodelExporter.java
```diff
@@ -32,7 +32,7 @@ public Resource exportResource(List<Language> languages) {
   public EPackage exportLanguage(Language language) {
     EPackage ePackage = EcoreFactory.eINSTANCE.createEPackage();
 
-    ePackage.setName(language.getName());
+    ePackage.setName(language.getName().replaceAll("[^\\p{Alnum}_]", "_"));
```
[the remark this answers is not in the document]
enikao: Yes, dots lead to invalid generated Java. I investigated the code of the `EcoreValidator`, and it checks essentially for valid Java identifiers.

emf/src/test/resources/properties.ecore
enikao: invalid, see above

emf/src/main/java/io/lionweb/lioncore/java/emf/EMFMetamodelExporter.java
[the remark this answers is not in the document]
ftomassetti: Makes sense, I would just add a comment as others could be tempted to remove it when we forget the reason for this

emf/src/test/resources/properties.ecore
ftomassetti: Understood, thank you

AtanasM98:
```
Caused by: org.eclipse.emf.ecore.resource.impl.ResourceSetImpl$1DiagnosticWrappedException: java.io.FileNotFoundException: 
C:\Work\lioncore-java\emf-builtins\build\resources\main\src\main\model\builtins.genmodel
(The system cannot find the path specified)
```
I'm having this problem. I think it may be connected to a missing gradle task or some task that sets up this resource isn't depended on.

enikao:
> ```
> Caused by: org.eclipse.emf.ecore.resource.impl.ResourceSetImpl$1DiagnosticWrappedException: java.io.FileNotFoundException: 
> C:\Work\lioncore-java\emf-builtins\build\resources\main\src\main\model\builtins.genmodel
> (The system cannot find the path specified)
> ```
> 
> I'm having this problem. I think it may be connected to a missing gradle task or some task that sets up this resource isn't depended on.

That should be fixed by now, could you try?

PR #116, opened 2023-11-10
Preparing for release 2023.1

ftomassetti:
* Adding changelog
* Adding specsVersion in artifacts names
* Moving to 0.2.0 to indicate a significant change and offer a more separate starting point for the changelog

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ The language's _nsURI_ is `http://lionweb.io/lionweb-java/emf/core/builtins/2023
 
 ## lionweb-java-2023.1-* - Version 0.2.0
 
-This is the first release based on the new coordinates, which indicate support for the specification of LionWeb released under version 2023.1.
+This is the first release based on the new artifact ids (which now include the specifications version), which indicate support for the specification of LionWeb released under version 2023.1.
 
 At this stage support for the M1 and M2 APIs is relatively complete. Support for annotations may need refinements. Some constraints may not yet been verified.
 Import and export from and to EMF is a work in progress.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -25,3 +25,12 @@ Contains im/exporters to convert models between LionCore &harr; Ecore and LionWe
 Some of LionCore's built-in elements have no direct representation in Ecore.
 This sub-project is an Eclipse project that defines an _EPackage_ `builtins` to host the equivalent elements in Ecore.
 The language's _nsURI_ is `http://lionweb.io/lionweb-java/emf/core/builtins/2023.1`.
+
+## Changelog
+
+## lionweb-java-2023.1-* - Version 0.2.0
+
+This is the first release based on the new coordinates, which indicate support for the specification of LionWeb released under version 2023.1.
```
dslmeinte, line 33: “coordinate**s**”? I'm not all too familiar with Maven parlance, so I could miss something. Could “package name” or something like that be more clear?

README.md
```diff
@@ -25,3 +25,12 @@ Contains im/exporters to convert models between LionCore &harr; Ecore and LionWe
...
+
+At this stage support for the M1 and M2 APIs is relatively complete. Support for annotations may need refinements. Some constraints may not yet been verified.
```
dslmeinte, line 35: This is a good format that I'll steal for `lionweb-typescript` :)

dslmeinte (changes requested):

README.md
ftomassetti: The term coordinates is indeed part of the Maven parlance (https://maven.apache.org/pom.html#Maven_Coordinates). It includes the groupId, the artifact id, and the version. I can change that to artifactIds, because we changed only those

README.md
ftomassetti: Great!

dslmeinte (approved):

PR #117, opened 2023-11-20
Steamlined low-level API

enikao:


core/src/main/java/io/lionweb/lioncore/java/language/IKeyed.java
```diff
@@ -1,9 +1,10 @@
 package io.lionweb.lioncore.java.language;
 
 /**
- * Implementation detail of the Java implementation. Not part of the official M3.
+ * Implementation detail of the Java implementation. Not part of the official M3. TODO: Is this
```
enikao, line 4: I think by now this is part of M3?

core/src/main/java/io/lionweb/lioncore/java/serialization/data/SerializedContainmentValue.java
```diff
@@ -1,18 +1,22 @@
...
 
-  public SerializedContainmentValue() {}
+  public SerializedContainmentValue() {
+    this.value = new ArrayList<>();
+  }
 
   public SerializedContainmentValue(MetaPointer metaPointer, List<String> value) {
     this.metaPointer = metaPointer;
-    this.value = value;
+    this.value = new ArrayList<>(value);
```
dslmeinte, line 19: Had to think why it would be useful to make a defensive copy here, but now I get it.

dslmeinte (approved):

core/src/main/java/io/lionweb/lioncore/java/serialization/data/SerializedContainmentValue.java
enikao: Nobody shall touch my internal data structure directly.

core/src/main/java/io/lionweb/lioncore/java/language/IKeyed.java
ftomassetti: Right, we should update the comment

core/src/main/java/io/lionweb/lioncore/java/serialization/JsonSerialization.java
```diff
@@ -161,9 +161,7 @@ public SerializedChunk serializeNodesToSerializationBlock(List<Node> nodes) {
               + " is not");
       registerLanguage(node.getConcept().getLanguage());
       UsedLanguage languageKeyVersion = UsedLanguage.fromLanguage(node.getConcept().getLanguage());
-      if (!serializationBlock.getLanguages().contains(languageKeyVersion)) {
-        serializationBlock.getLanguages().add(languageKeyVersion);
-      }
+      serializationBlock.addLanguage(languageKeyVersion);
```
ftomassetti, line 164: Don't we want to prevent accidental double inclusions?

core/src/main/java/io/lionweb/lioncore/java/serialization/data/SerializedChunk.java
```diff
@@ -13,7 +13,7 @@ public class SerializedChunk {
   private final Map<String, SerializedClassifierInstance> classifierInstancesByID = new HashMap<>();
 
   private String serializationFormatVersion;
-  private final List<UsedLanguage> languages = new ArrayList<>();
+  private final Set<UsedLanguage> languages = new LinkedHashSet<>();
```
ftomassetti, line 16: I would not use a set here for two reasons: 1) Preserving the order 2) Load faithfully files containing twice the same language

core/src/main/java/io/lionweb/lioncore/java/serialization/data/SerializedChunk.java
```diff
@@ -47,11 +47,11 @@ public void addLanguage(UsedLanguage language) {
   }
 
   public Map<String, SerializedClassifierInstance> getClassifierInstancesByID() {
-    return classifierInstancesByID;
+    return Collections.unmodifiableMap(classifierInstancesByID);
   }
 
-  public List<UsedLanguage> getLanguages() {
-    return languages;
+  public Set<UsedLanguage> getLanguages() {
```
ftomassetti, line 53: I would return a List here

core/src/main/java/io/lionweb/lioncore/java/serialization/data/UsedLanguage.java
```diff
@@ -25,6 +25,13 @@ public static UsedLanguage fromLanguage(@Nonnull Language language) {
     return new UsedLanguage(language.getKey(), language.getVersion());
   }
 
+  public static UsedLanguage fromMetaPointer(@Nonnull MetaPointer metaPointer) {
+    Objects.requireNonNull(metaPointer, "metaPointer parameter should not be null");
+    Objects.requireNonNull(metaPointer.getLanguage(), "metaPointer language should not be null");
```
ftomassetti, line 30: Should we have a method to validate a metaPointer and have the null checks there?

ftomassetti (changes requested):
All looks good to me, besides the use of a set in SerializedChunk. I think that class should represent what is on file, even if incorrect and therefore it is useful for it to preserve the order of elements and even duplicates

core/src/main/java/io/lionweb/lioncore/java/serialization/data/SerializedChunk.java
dslmeinte: `LinkedHashSet` #forTheWin !

ftomassetti: I would still think that in SerializedChunk it makes sense to allow for duplicates. If we don't we would not be able to deserialize and reserialize a file and get the same result

dslmeinte: I agree (even if `LinkedHashSet` preserves order modulo removals): we shouldn't have any "magic".

core/src/main/java/io/lionweb/lioncore/java/serialization/data/UsedLanguage.java
enikao: I wrote this symmetric to `fromLanguage()`, didn't think too much about it.
In theory, we could work with a MetaPointer that misses its `key`.
I don't have a strong opinion, what do you prefer?

ftomassetti (approved):
+1

PR #118, opened 2023-12-12
Simplify internal representation of serialized nodes + re-order JSON serialization

dslmeinte:


README.md
```diff
@@ -34,3 +34,9 @@ This is the first release based on the new artifact ids (which now include the s
 
 At this stage support for the M1 and M2 APIs is relatively complete. Support for annotations may need refinements. Some constraints may not yet been verified.
 Import and export from and to EMF is a work in progress.
+
+### Version 0.2.1
```
ftomassetti, line 38: I think it could make sense to have a separate Changelog file for this

core/src/main/java/io/lionweb/lioncore/java/serialization/JsonSerialization.java
```diff
@@ -414,30 +414,13 @@ private List<SerializedClassifierInstance> sortLeavesFirst(
...
-              || sortedList.stream()
-                  .anyMatch(
-                      sn ->
-                          Objects.equals(
-                              sn.getID(), serializedAnnotationInstance.getParentNodeID()))) {
-            sortedList.add(n);
-            nodesToSort.remove(i);
-            i--;
-          }
+        SerializedClassifierInstance node = nodesToSort.get(i);
```
ftomassetti, line 417: should we call the variable `classifierInstance` or `instance` or `ci`?

core/src/main/java/io/lionweb/lioncore/java/serialization/JsonSerialization.java
```diff
@@ -496,19 +479,16 @@ private List<ClassifierInstance<?>> deserializeClassifierInstances(
             new MapBasedResolver(deserializedByID), this.instanceResolver);
     serializedClassifierInstances.stream()
         .forEach(
-            n -> {
+            node -> {
```
ftomassetti, line 482: should we call the variable `classifierInstance` or `instance` or `ci`?

core/src/main/java/io/lionweb/lioncore/java/serialization/JsonSerialization.java
```diff
@@ -496,19 +479,16 @@ private List<ClassifierInstance<?>> deserializeClassifierInstances(
...
-                SerializedAnnotationInstance serializedAnnotationInstance =
-                    (SerializedAnnotationInstance) n;
-                if (serializedAnnotationInstance == null) {
+                if (node == null) {
                   throw new IllegalStateException(
                       "Dangling annotation instance found (annotated node is null). ");
                 }
-                Node annotatedNode =
-                    (Node) deserializedByID.get(serializedAnnotationInstance.getParentNodeID());
+                Node annotatedNode = (Node) deserializedByID.get(node.getParentNodeID());
```
ftomassetti, line 491: Can an annotations have children? If so the parent of an annotation could be another annotation and this could fail

core/src/main/java/io/lionweb/lioncore/java/serialization/data/SerializedClassifierInstance.java
```diff
@@ -3,22 +3,23 @@
...
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 /**
  * Lower level representation of a Classifier Instance (either a Node or an AnnotationInstance)
- * which is used to load broken classifier instances during serialization.
+ * which is used to load classifier instances during serialization.
```
ftomassetti, line 12: I think that the broken term here makes sense as we can use these classes also to load nodes that are not structurally valid (like nodes that have themselves as parent)

emf/src/test/resources/properties-instance.json
```diff
@@ -4,6 +4,11 @@
   "nodes": [
     {
       "id": "file_properties-parser-examples-example1-props",
+      "classifier": {
```
ftomassetti, line 7: Is this necessary to change the file? Shouldn't the test pass anyway?

emf/src/test/resources/properties-language.json
```diff
@@ -9,6 +9,11 @@
   "nodes": [
     {
       "id": "io_lionweb_Properties",
+      "classifier": {
```
ftomassetti, line 12: As for the other file, I would consider leaving the file as-is, to demonstrate that the test should pass anyway and it is not dependent on the order of keys

ftomassetti (approved):
Looks good to me, with minor suggestions for coherence on the names of variables

README.md
dslmeinte: Ha, I thought we decided to not have separate changelog files. Also: the `##` heading says "Changelog".

core/src/main/java/io/lionweb/lioncore/java/serialization/JsonSerialization.java
dslmeinte: I'd rather rename the `SerializedClassifierInstance` to `SerializedNode` (and so `node` &rarr; `serializedNode`), actually.

core/src/main/java/io/lionweb/lioncore/java/serialization/JsonSerialization.java
dslmeinte: (see previous comment)

README.md
ftomassetti: Sorry, my bad

core/src/main/java/io/lionweb/lioncore/java/serialization/JsonSerialization.java
ftomassetti: I am confused about the term "node", as it could mean both node as in proper member of the AST (so no annotation instances) and node as in proper member of the AST or annotation instance. I think it would be useful to have separate terms for these two concepts

core/src/main/java/io/lionweb/lioncore/java/serialization/JsonSerialization.java
dslmeinte: I don't know if annotations can have children. (This change doesn't impact that, though.) I'll also check the TS-impl. for what we do there.

core/src/main/java/io/lionweb/lioncore/java/serialization/data/SerializedClassifierInstance.java
dslmeinte: Yes, but they're not _only_ used for broken classifiers, right? If so, we can state explicitly that this class is also usable to load broken serialized nodes.

emf/src/test/resources/properties-instance.json
dslmeinte: The test passes anyway, but I like consistency :)

emf/src/test/resources/properties-language.json
dslmeinte: Shouldn't we make a separate test case (and file) for that, then?

README.md
dslmeinte: No worries :D

core/src/main/java/io/lionweb/lioncore/java/serialization/JsonSerialization.java
dslmeinte: In TS and C# code I tend to use the names `node` and `serializedNode` – can we do that here as well?

core/src/main/java/io/lionweb/lioncore/java/serialization/JsonSerialization.java
dslmeinte: An `Annotation` is a `Classifier` so can own `Feature`s. It can't own another `Annotation`, only reference them through its `extends` feature. So, I'd say the code as it is, is OK. (In the TS-impl. it wouldn't be a problem anyway, because of the way that parent-child-relations are installed.)

core/src/main/java/io/lionweb/lioncore/java/serialization/JsonSerialization.java
ftomassetti: I am not super-convinced about this point, but perhaps we can revisit it later and keep things as they are for the sake of getting this PR over the line and the release done quick;y

core/src/main/java/io/lionweb/lioncore/java/serialization/data/SerializedClassifierInstance.java
ftomassetti: You are right, I added a suggestion

emf/src/test/resources/properties-language.json
ftomassetti: Yes, I think so. I will mark this as resolved then

core/src/main/java/io/lionweb/lioncore/java/serialization/JsonSerialization.java
ftomassetti: Marking this as resolved

core/src/main/java/io/lionweb/lioncore/java/serialization/data/SerializedClassifierInstance.java
ftomassetti: I committed my own suggestion, to get this one over the line

PR #119, opened 2024-01-02
Align order of language meta-pointers

dslmeinte:
+ fix lint issue


ftomassetti:
Thank you @dslmeinte . Please let me know if you need a release

dslmeinte:
> Thank you @dslmeinte . Please let me know if you need a release

That would be good, although I don't know whether you have other work in the pipeline that would be good to release in combination.

ftomassetti:
@dslmeinte I just made a release (0.2.3) it should now propagate to Maven Central. Please let me know if you run into issues with it, as they could be due to the new release process

PR #121, opened 2024-01-02
Setup advanced publish gradle plugin

### LINKED ISSUE #120: Adopt com.vanniktech.maven.publish to simplify the release process
See https://github.com/Strumenta/antlr-kotlin/blob/master/build.gradle.kts for reference. It permits to auto close and release the sonatype repository, making the release process fully automatic

ftomassetti:
Fix #120 

This PR:
1. Simplifies slightly the configuration of the build
2. Automatize the release process, removing the manual step of logging into sonatype, close the repository and then release it

I have tested it by publishing the artifacts locally, but I am not sure how to test it differently than by doing a real release after merging this PR.

I have previously used this plugin in https://github.com/Strumenta/antlr-kotlin

I think that most of the code in mavenPublishing could be centralized under the `buildSrc` directory, but I am not sure if we can use that while having a mix of groovy and kotlin scripts. At this moment we cannot fully move to kotlin because of EMF stuff that requires some groovy code.

Given that JDK 8 was not supporting the gradle plugin introduced in this PR, I also updated the list of JDK supported for the build process. I dropped JDK 8 (which could already not run the javadoc task) and added JDK 21 (a new LTS). This required updating gradle. We still generate artifacts compatible with JDK 8.


ftomassetti:
Converting to draft as the plugin seems to require Java 11, for the build process. This makes the build fail, even if it should not be a problem as the code produced should still be Java 8 compatible. Still, the build should be fixed.

dslmeinte (approved):

ftomassetti:
Updated MWE and Xtext so that tests on JDK 21 pass now

PR #126, opened 2024-02-08
Avoid duplicate features in case of inheritance

### LINKED ISSUE #125: Bug: duplicates in features
Methods like `node.concept.inheritedFeatures()`, `allProperties` can return twice the same feature, if it has been inherited twice.

ftomassetti:
Fix #125

### DIFF
--- a/core/src/main/java/io/lionweb/lioncore/java/language/Annotation.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/language/Annotation.java
@@ -111,10 +111,10 @@ public class Annotation extends Classifier<Annotation> {
   public List<Feature<?>> inheritedFeatures() {
     List<Feature<?>> result = new LinkedList<>();
     if (this.getExtendedAnnotation() != null) {
-      result.addAll(this.getExtendedAnnotation().allFeatures());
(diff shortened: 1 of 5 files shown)


ftomassetti:
Right, I added two unit tests. One of them is also checking for inheritance of features in the presence of inheritance loops.
This actually helped finding a bug. Thank you @dslmeinte !
One challenge is due to the fact that we want to produce features in the inheritance orders (first the features declared in the element, then the one inherited from direct ancestors and so on).

ftomassetti:
I am merging and making a new release

dslmeinte:
> Right, I added two unit tests. One of them is also checking for inheritance of features in the presence of inheritance loops. This actually helped finding a bug. Thank you @dslmeinte ! One challenge is due to the fact that we want to produce features in the inheritance orders (first the features declared in the element, then the one inherited from direct ancestors and so on).

I remember running into the same thing for TypeScript. There, I used a `flatMapNonFollowingCycles` function that does a tree traversal (with a function parameter for the next nodes to follow, and one to append to the result). With that, `allFeatures` and similar functions become 1-liners.

PR #127, opened 2024-02-09
Rename Model into Partition and implement Node.getRoot

ftomassetti:
The first item is just a simple refactoring of something I noticed while working on Node.getRoot.

For the second point, I wrote two tests: one for the "normal" case, and one to handle circular hierarchies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,23 @@
+# Changelog
+
+## lionweb-java-2023.1-* - Version 0.2.0
+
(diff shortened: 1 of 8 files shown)


dslmeinte (approved):
Do you already have a changelog for `lionweb-java`? Would be good to have, since e.g. this PR introduces a breaking change.

ftomassetti:
Good point @dslmeinte I have updated the Changelog and moved it to a separate file

PR #128, opened 2024-02-13
Correct serialization/deserialization of nullable boolean values

ftomassetti:
Nullable boolean values were treating null and false as interchangeable values, so that serializing and deserializing one would get back a different value.

For properties of type boolean which are _optional_ we want to distinguish _false_ and _null_, so the code has been corrected and tests have been added.

### DIFF
--- a/core/src/test/java/io/lionweb/lioncore/java/model/impl/DynamicNodeTest.java
+++ b/core/src/test/java/io/lionweb/lioncore/java/model/impl/DynamicNodeTest.java
@@ -3,10 +3,7 @@ package io.lionweb.lioncore.java.model.impl;
 import static org.junit.Assert.*;
 
 import com.google.gson.JsonArray;
-import io.lionweb.lioncore.java.language.Annotation;
-import io.lionweb.lioncore.java.language.Concept;
-import io.lionweb.lioncore.java.language.Containment;
-import io.lionweb.lioncore.java.language.Language;
+import io.lionweb.lioncore.java.language.*;
 import io.lionweb.lioncore.java.model.AnnotationInstance;
 import io.lionweb.lioncore.java.serialization.MyNodeWithProperties;
 import java.util.Arrays;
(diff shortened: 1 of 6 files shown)


core/src/test/java/io/lionweb/lioncore/java/serialization/MyNodeWithProperties.java
```diff
@@ -33,8 +33,13 @@ public MyNodeWithProperties(String id) {
     super(id, CONCEPT);
   }
 
-  public boolean getP1() {
-    return (boolean) this.getPropertyValueByName("p1");
+  public Boolean getP1() {
+    Object value = this.getPropertyValueByName("p1");
+    if (value == null) {
```
dslmeinte, line 38: In test code, so not so important, but I find this logic better readable when using a ternary operator.

ftomassetti: Fixed

core/src/test/java/io/lionweb/lioncore/java/model/impl/DynamicNodeTest.java
```diff
@@ -173,4 +170,64 @@ public void getRootCircularHierarchy() {
...
+    Concept a = new Concept(l, "A", "a-id", "a-key");
+    a.addFeature(Property.createOptional("foo", LionCoreBuiltins.getBoolean()));
+    DynamicNode n1 = new DynamicNode("n1", a);
+
+    assertEquals(null, n1.getPropertyValueByName("foo"));
+    n1.setPropertyValueByName("foo", null);
+    assertEquals(null, n1.getPropertyValueByName("foo"));
+  }
+
+  @Test
```
enikao, line 222: I think we're missing two tests:

a) setting `null` to a non-nullable property
b) setting `null` to a nullable property that currently has a value

ftomassetti:
Thank you @dslmeinte !

core/src/test/java/io/lionweb/lioncore/java/serialization/MyNodeWithProperties.java
```diff
@@ -33,8 +33,9 @@ public MyNodeWithProperties(String id) {
     super(id, CONCEPT);
   }
 
-  public boolean getP1() {
-    return (boolean) this.getPropertyValueByName("p1");
+  public Boolean getP1() {
+    Object value = this.getPropertyValueByName("p1");
+    return value == null ? null : (Boolean) value;
```
enikao, line 38: Is this check actually needed? I think we can cast `null` to `Boolean`

ftomassetti: I will check that. It is such a long time that I do not use Java!

core/src/test/java/io/lionweb/lioncore/java/model/impl/DynamicNodeTest.java
ftomassetti: I will add these tests, thank you

core/src/test/java/io/lionweb/lioncore/java/serialization/MyNodeWithProperties.java
ftomassetti: You are right, I fixed that in master as I had already merged this

core/src/test/java/io/lionweb/lioncore/java/model/impl/DynamicNodeTest.java
ftomassetti: Tests added directly in master, as I had already merged this

PR #129, opened 2024-02-14
Check that DataType.id is not null during serialization

ftomassetti:
Here the intent is just producing clearer error messages

### DIFF
--- a/core/src/main/java/io/lionweb/lioncore/java/serialization/JsonSerialization.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/serialization/JsonSerialization.java
@@ -377,7 +377,10 @@ public class JsonSerialization {
   // Deserialization - Private
   //
 
-  private String serializePropertyValue(DataType dataType, Object value) {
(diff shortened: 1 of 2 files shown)


dslmeinte (approved):

core/src/main/java/io/lionweb/lioncore/java/serialization/JsonSerialization.java
```diff
@@ -377,7 +377,10 @@ public List<Node> deserializeToNodes(InputStream inputStream) {
   // Deserialization - Private
   //
 
-  private String serializePropertyValue(DataType dataType, Object value) {
+  private String serializePropertyValue(@Nonnull DataType dataType, Object value) {
```
enikao, line 380: Why value not marked `@Nullable`?

ftomassetti: You are right, I have now marked it

enikao (approved):

ftomassetti:
Thank you @enikao and @dslmeinte  for the revisions

PR #132, opened 2024-02-20
Support deserialization of partial trees

ftomassetti:
Until now we were expecting to receive entire trees. In other words to receive some node with a null parent and other nodes directly or indirectly connected to such roots. In this way we were able to unserialize the tree properly, setting the parents for all nodes.

However, with support for proper partitions being introduced, we may want to ask for specific root nodes (or smaller trees) and not always to get the whole partition. So we may need to be able to unserialize trees that refer to parents we did not receive. We therefore ask support for this. In that case the parents not transmitted in the chunk will be set to null. 

In the future we may want to specify some form of NodeResolver.

### DIFF
--- a/core/src/main/java/io/lionweb/lioncore/java/serialization/JsonSerialization.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/serialization/JsonSerialization.java
@@ -22,6 +22,16 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 /**
+ * When deserializing a tree, we either extract an entire partitions (and perhaps all the referred
+ * partitions) or we will get references to nodes (for example, parents or ancestors) outside of the
+ * scope of the tree extracted. This policy specifies what we do with such references.
+ */
+enum UnknownNodePolicy {
+  NULL_REFERENCES,
+  THROW_ERROR
+}
+
+/**
  * This class is responsible for deserializing models.
  *
  * <p>The deserialization of each node _requires_ the deserializer to be able to resolve the Concept
@@ -100,6 +110,8 @@ public class JsonSerialization {
 
   private final LocalClassifierInstanceResolver instanceResolver;
 
+  private UnknownNodePolicy unknownNodePolicy = UnknownNodePolicy.THROW_ERROR;
+
   private JsonSerialization() {
     // prevent public access
     classifierResolver = new ClassifierResolver();
@@ -131,6 +143,15 @@ public class JsonSerialization {
   public void enableDynamicNodes() {
     instantiator.enableDynamicNodes();
     primitiveValuesSerialization.enableDynamicNodes();
+  }
+
+  public @Nonnull UnknownNodePolicy getUnknownNodePolicy() {
+    return this.unknownNodePolicy;
+  }
+
+  public void setUnknownNodePolicy(@Nonnull UnknownNodePolicy unknownNodePolicy) {
+    Objects.requireNonNull(unknownNodePolicy);
+    this.unknownNodePolicy = unknownNodePolicy;
   }
 
   //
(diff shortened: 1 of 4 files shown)


enikao:
I think we can make a clear distinction: The lowlevel stuff (`SerializedClassifierInstance` etc.) should work just fine with unknown node ids. As soon as we cross over to the higher-level API, they _must_ be resolved -- otherwise we cannot hold any of the semantic guarantees.

I only had a glimpse on this PR, so I don't know how it compares to this idea.

ftomassetti:
> I think we can make a clear distinction: The lowlevel stuff (`SerializedClassifierInstance` etc.) should work just fine with unknown node ids. 

Yes, I agree and this is how the code works

> As soon as we cross over to the higher-level API, they _must_ be resolved -- otherwise we cannot hold any of the semantic guarantees.

Well, yes, but that means that if I am interested in retrieve _any_ node, I must get all the nodes connected and referred on the Client side, otherwise I cannot unserialize the tree.

So if I have a large partition, containing many trees directly under it, when I get any node within that partition I also need to get (or already have somehow on the client) all the parents, up to the partition, and then all the children and descendant of the partition. So I need to get a lot of stuff.

I suggest having the possibility to ask for subtrees and get the subtree that is of interest "cutting away" the parent and their ancestors, but doing that only explicitly (i.e., passing a flag indicating this is the behavior expected). When not setting the flag asking for anything but an entire partition just fails, as we do not have the nodes that we need to refer from one subtree

ftomassetti:
Perhaps we can use some form of proxy or placeholder, but I think it would be important to have a way to retrieve and unserialize something smaller than an entire partition. And if we decide to forbid that, then we should redefine the bulk APIs to specify that only partitions can be retrieve, and not anything smaller.

enikao:
Bulk API explicitly supports partial trees, so your approach makes sense.

Regarding the implementation, it looks good.
The only think I'm not so sure about is the API with the boolean flag. We already see that you had to change lots of places. Any future change (e.g. providing a NodeResolver) would involve at least as many changes. Maybe we should introduce a builder or config to encapsulate that?

ftomassetti:
> The only think I'm not so sure about is the API with the boolean flag. We already see that you had to change lots of places. Any future change (e.g. providing a NodeResolver) would involve at least as many changes. Maybe we should introduce a builder or config to encapsulate that?

We could make this a setting of JsonSerialization. It would probably make sense. I will refactor the code accordingly.

core/src/main/java/io/lionweb/lioncore/java/serialization/JsonSerialization.java
```diff
@@ -21,6 +21,16 @@
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
+/**
+ * When deserializing a tree, we either extract an entire partitions (and perhaps all the referred
+ * partitions) or we will get references to nodes (for example, parents or ancestors) outside of the
+ * scope of the tree extracted. This policy specifies what we do with such references.
+ */
+enum UnknownNodePolicy {
```
enikao, line 29: This policy is only about unknown _parents_, correct? Unknown annotations / children / reference targets are not affected. We should reflect that in the name.

enikao:
By the way, we _can_ have results from bulk API that don't contain children -- if we set the maxDepth parameter. We might want to deal with that too (maybe in a different PR).

core/src/main/java/io/lionweb/lioncore/java/serialization/JsonSerialization.java
ftomassetti: Right, I will do that

ftomassetti:
> By the way, we _can_ have results from bulk API that don't contain children -- if we set the maxDepth parameter. We might want to deal with that too (maybe in a different PR).

Good point. I will create an issue for that

ftomassetti:
I have created a follow-up issue for dealing with missing children, have renamed the enum to UnknownParentPolicy, and rebased on master

enikao (approved):

ftomassetti:
Thank you @enikao !

PR #137, opened 2024-04-11
Representing enumeration values

ftomassetti:
When a property is of type enumeration, we had not a clear expectation on how the value should be represented. In different places we were expecting:
* an EnumerationLiteral
* a String
* a DynamicEnumerationValue

With this PR we define the interface EnumerationValue (and the class EnumerationValueImpl, previously called DynamicEnumerationValue). The standard approach should be to use this to represent enumeration value. This will make it clear that we are not representing just a string or the definition a new enumeration literal but representing an instance of an enumeration corresponding to a certain literal.

Users can still decide to represent enumeration values in other ways, if they wish so, but in that case they would need to provide primitive serializers and deserializers. If they instead use the EnumerationValue serialization and deserialization will work out of the box.


core/src/main/java/io/lionweb/lioncore/java/serialization/PrimitiveValuesSerialization.java
```diff
@@ -152,11 +162,20 @@ public String serialize(@Nonnull String primitiveTypeID, @Nullable Object value)
...
-      // (and not the ID)
+      // (and not the ID).
       // This is at least the default behavior, but the user can register specialized
-      // primitiveSerializers,
-      // if a different behavior is needed
-      return ((EnumerationLiteral) value).getKey();
+      // primitiveSerializers, if a different behavior is needed
+      EnumerationLiteral enumerationLiteral;
+      if (value instanceof EnumerationValue) {
+        enumerationLiteral = ((EnumerationValue) value).getEnumerationLiteral();
```
dslmeinte, line 170: Why not `return ((EnumerationValue) value).getEnumerationLiteral().getKey();` here, then? The `else`-branch just `throw`s.

dslmeinte (approved):

enikao:
Do we still have tests that use actual java enums?

ftomassetti:
> Do we still have tests that use actual java enums?

I am not sure if we ever had such tests. However this comment gives me the idea that we may have support for using enum values to represent values of properties of Enumeration type. In that case, if we had an enum instance we would not not from that to which Enumeration and EnumerationLiteral it would be connected. However, knowing the property to which it is assigned we could get the Enumeration from there and then look at the EnumerationLiteral with the same name as the enum instance. Do you think it would make sense to add support for that as part of this PR?

enikao:
> Do you think it would make sense to add support for that as part of this PR?

Yes, as this PR claims to bring the flexibility to do both. Then we should also test it.

ftomassetti:
> Yes, as this PR claims to bring the flexibility to do both. Then we should also test it.

This is not what I meant initially, but I think it is a good idea so I added support for using Enum instances to represent instances of EnumerationLiteral and added some tests for serialization and deserialization of such values

core/src/main/java/io/lionweb/lioncore/java/serialization/PrimitiveValuesSerialization.java
```diff
@@ -173,7 +188,74 @@ public String serialize(@Nonnull String primitiveTypeID, @Nullable Object value)
...
     return enumerationsByID.containsKey(primitiveTypeID);
   }
+
+  public static <E extends Enum<E>> PrimitiveSerializer<E> serializerFor(
+      Class<E> enumClass, Enumeration enumeration) {
+    return value -> {
+      String enumerationLiteralName = value.name();
+      Optional<EnumerationLiteral> enumerationLiteral =
+          enumeration.getLiterals().stream()
+              .filter(l -> l.getName().equals(enumerationLiteralName))
```
enikao, line 207: We identify enum literals by key, not by name.

ftomassetti: Yes, we use the key for serialization and deserialization (see for example the return statement of this method). Here however we are trying to map an enumerationLiteral to a particular enum constant from an Enum class. In that case we use the name to pair them.

In other words, let's say we have an enum class MyEnum:

```
enum MyEnum { FOO, BAR }
``` 

In this method we register a primitive serializer that associates instances of MyEnum with a certain Enumeration. In this serializer we will receive an instance of MyEnum, for example `FOO`. We will then look for an enumeration literal named `FOO` among the literals of the enumeration associated with the enum class. But then, once we have the enumeration literal, we use its key for producing the serialization value.

enikao: ok, I get it. Names must always be unique in Java, so it should work.

PR #139, opened 2024-04-18
Minor fixes around DynamicNode and serialization

ftomassetti:
These are small changes mostly in DynamicNode.

* We permit to modify `id` and `concept` after instantiation
* We implement `getContainmentFeature`

We also solve a problem with deserializing of null parents.


dslmeinte (approved):

ftomassetti:
Thank you @dslmeinte !

PR #148, opened 2024-06-06
Revisit node class

ftomassetti:
The refactoring mostly focuses on adding utility methods for dealing with references and containments.

In doing so, some reorganization of methods within related classes (Node, DynamicClassifierInstance, AbstractClassifierInstance) emerged, also consequently to the recent introduction of AbstractClassifierInstance.

* In Node we removed `getConcept` as `getClassifier` is specialized to return a `Concept` for `Node`, so `getConcept` was redundant at this stage. This change alone is the one affecting most of the files. We could potentially keep `getConcept` and deprecating it, but given the small user base, I would just go ahead and remove it

### DIFF
--- a/core/src/main/java/io/lionweb/lioncore/java/language/Annotation.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/language/Annotation.java
@@ -120,7 +120,7 @@ public class Annotation extends Classifier<Annotation> {
   }
 
   @Override
-  public Concept getConcept() {
+  public Concept getClassifier() {
     return LionCore.getAnnotation();
   }
 }
--- a/core/src/main/java/io/lionweb/lioncore/java/language/Concept.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/language/Concept.java
@@ -121,7 +121,7 @@ public class Concept extends Classifier<Concept> {
   }
 
   @Override
-  public Concept getConcept() {
+  public Concept getClassifier() {
     return LionCore.getConcept();
   }
 }
--- a/core/src/main/java/io/lionweb/lioncore/java/language/Containment.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/language/Containment.java
@@ -107,7 +107,7 @@ public class Containment extends Link<Containment> {
   }
 
   @Override
-  public Concept getConcept() {
+  public Concept getClassifier() {
     return LionCore.getContainment();
   }
 }
--- a/core/src/main/java/io/lionweb/lioncore/java/language/Enumeration.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/language/Enumeration.java
@@ -30,7 +30,7 @@ public class Enumeration extends DataType<Enumeration> implements NamespaceProvi
   }
 
   @Override
-  public Concept getConcept() {
+  public Concept getClassifier() {
     return LionCore.getEnumeration();
   }
 }
--- a/core/src/main/java/io/lionweb/lioncore/java/language/EnumerationLiteral.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/language/EnumerationLiteral.java
@@ -52,7 +52,7 @@ public class EnumerationLiteral extends M3Node<EnumerationLiteral>
   }
 
   @Override
-  public Concept getConcept() {
+  public Concept getClassifier() {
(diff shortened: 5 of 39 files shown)


ftomassetti:
This PR has a limited impact, but it is "large" in the sense that it touches 50 files. This makes it time-consuming to review. It is mostly about moving methods around and adding a few convenient methods to handle references (similar to methods we already had for containments). 

I am planning to go ahead and merge it. So, I could then work on other things (like increasing test coverage or supporting recent changes in the specs).
Anyone opposed?

enikao:
A very generic comment: I find our pretty big API surface dangerous in the sense that we limit our future possibilities. I'd aim at a minimal API in the core classes, and maybe some convenience methods around it. (Extension methods would simplify this, but they aren't available in Java.)

Case in point: When I tried to change the implementation to stream processing, I had to adjust every variant of the methods in `NodeSerializer`; some of them would be pretty hard to adjust at all.

Another issue with that many methods in Node API are potential clashes with generated domain methods: If I used this API as a base for generated classes, I had to create special handling for any feature called `Classifier`, `ReferenceValues`, etc.
Personally, I liked EMF's approach to this -- having `e` prefixes for all API methods, for example `eGet()`.

ftomassetti:
> A very generic comment: I find our pretty big API surface dangerous in the sense that we limit our future possibilities. I'd aim at a minimal API in the core classes, and maybe some convenience methods around it. (Extension methods would simplify this, but they aren't available in Java.)

I agree that having extension methods would be great. My rationale for adding convenience methods is that these methods are needed in practice to use the library conveniently. If we do not write such methods as part of the library, users will need to add them, in cases where they can, while in other cases they would be stuck. This PR initially started because I was attempting to do something with references and I then realize I was missing APIs (probably because originally I thought that we did not support multiple references, so I did not add methods I added to deal with lists as I did for containments).

For the convenience methods that can be implemented externally, it would lead to more bugs and headaches because users would not know about LionWeb as much as we do, while we are in the best position to write them correctly. Ideally, as many methods as possible should be implemented as default methods in interfaces so that we can simplify the life of implementers that, for some reason, want not to use the standard base classes (like your streaming use cases).


> Another issue with that many methods in Node API are potential clashes with generated domain methods: If I used this API as a base for generated classes, I had to create special handling for any feature called Classifier, ReferenceValues, etc.
Personally, I liked EMF's approach to this -- having e prefixes for all API methods, for example eGet().

Maybe we could prefix a bunch of methods with the `lw` prefixes. It would also make some code I have more obvious as in Kolasu we end up having very similar method names and distinguishing them is not obvious. Should we open an issue about this?

enikao:
> For the convenience methods that can be implemented externally, it would lead to more bugs and headaches because users would not know about LionWeb as much as we do, while we are in the best position to write them correctly. Ideally, as many methods as possible should be implemented as default methods in interfaces so that we can simplify the life of implementers that, for some reason, want not to use the standard base classes (like your streaming use cases).

We could implement them as static methods in utility classes. With `import static`, we could get close to extension methods: not `myNode.addReferences(r1, r2)`, but at least `addReferences(myNode, r1, r2)`. It would still be under our quality control, and somewhat explorable.

> Maybe we could prefix a bunch of methods with the `lw` prefixes. It would also make some code I have more obvious as in Kolasu we end up having very similar method names and distinguishing them is not obvious. Should we open an issue about this?

Good idea.

ftomassetti:
> We could implement them as static methods in utility classes. With import static, we could get close to extension methods: not myNode.addReferences(r1, r2), but at least addReferences(myNode, r1, r2). It would still be under our quality control, and somewhat explorable.

I created `ClassifierInstanceUtils` and moved as many methods as possible there. This should reduce the API provided by Node and ClassifierInstance

For prefixing remaining methods with `lw`, I would open a separate issue and discuss it as a follow up. Would this work for you?

ftomassetti:
We the introduction of `ClassifierInstanceUtils` (as suggested by @enikao ) we significantly reduced the number of methods exposed by Node and ClassifierInstance. This also made the introduction of `AbstractNode` unnecessary.
In LionWeb Kotlin (not yet open-source) I introduced extension methods, wrapping the static methods provided by `ClassifierInstanceUtils`. It would make sense to release it, as soon as it stabilizes.

PR #152, opened 2024-06-14
Support multiple references of the same proxied node in the same chunk

ftomassetti:
Currently if we have two or more references to the same _external_ node (i.e., a node not present in the chunk) then we attempt to create multiple instances of `ProxyNode`. The moment we try to create the second, an exception is thrown.
This PR solves this bug and adds a test to verify that.


dslmeinte (approved):

ftomassetti:
Thank you @dslmeinte !

PR #154, opened 2024-06-27
Exposing SerializedJsonComparisonUtils

ftomassetti:
Exposing `SerializedJsonComparisonUtils` for the reason explained in https://github.com/LionWeb-io/specification/issues/283

In practice, the format currently uses lists for things for which the order is not meaningful, so to verify that two chunks are equivalent require specific logic.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -26,7 +26,7 @@ val javadocConfig by configurations.creating {
 
 dependencies {
     // Use JUnit test framework.
-    implementation(libs.junit)
+    testImplementation(libs.junit)
 
     // This dependency is exported to consumers, that is to say found on their compile classpath.
     api("org.apache.commons:commons-math3:3.6.1")
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -26,7 +26,7 @@ val javadocConfig by configurations.creating {
 
 dependencies {
     // Use JUnit test framework.
-    testImplementation(libs.junit)
+    implementation(libs.junit)
 
     // This dependency is exported to consumers, that is to say found on their compile classpath.
     api("org.apache.commons:commons-math3:3.6.1")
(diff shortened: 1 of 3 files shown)


core/build.gradle.kts
```diff
@@ -26,7 +26,7 @@ val javadocConfig by configurations.creating {
 
 dependencies {
     // Use JUnit test framework.
-    testImplementation(libs.junit)
```
enikao, line 29: Doesn't that mean that we have to have JUnit in our production environment?

ftomassetti: I would say no, because this is a `testImplementation` dependency, so it is needed only by tests, which are not packaged, published and reused in the production environment

ftomassetti: Right, sorry, I did not notice that the comment was associated with the old version of the code and not the new one.

So, yes, you are right. We are making junit a dependency of the library as we use assert methods in this utility.

Let me see if we can avoid doing that.

ftomassetti:
I removed the dependency to JUnit. @enikao do you think this can now be merged?

core/src/main/java/io/lionweb/lioncore/java/serialization/SerializedJsonComparisonUtils.java
```diff
@@ -12,11 +9,11 @@
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
-class SerializedJsonComparisonUtils {
+public class SerializedJsonComparisonUtils {
 
   private SerializedJsonComparisonUtils() {}
 
-  static void assertEquivalentLionWebJson(JsonObject expected, JsonObject actual) {
+  public static void assertEquivalentLionWebJson(JsonObject expected, JsonObject actual) {
```
enikao, line 16: What about making this _not_ static? Then we could override `assertEquals()` and `fail()` in subclasses if we wanted to handle them differently. (These two methods would need to be `protected`, of course.)

ftomassetti: For the time being, I do not see a situation where we may want to subclass this class, so for simplicity I would keep those methods as static until there is a specific use case

PR #156, opened 2024-06-28
Move addAnnotation and removeAnnotation to ClassifierInstance

### LINKED ISSUE #153: Annotations on non-dynamic nodes
Currently, it is only possible to add annotations on dynamic nodes because the _annotations_ collection is an immutable view and the _addAnnotation_ method is only available on _AbstractClassifierInstance_ objects.

ftomassetti:
Move to ClassifierInstance annotations specific methods previously only available on AbstractClassifierInstance
Fix #153

### DIFF
--- a/core/src/main/java/io/lionweb/lioncore/java/model/ClassifierInstance.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/model/ClassifierInstance.java
@@ -1,13 +1,35 @@
 package io.lionweb.lioncore.java.model;
 
+import io.lionweb.lioncore.java.language.Annotation;
 import io.lionweb.lioncore.java.language.Classifier;
 import io.lionweb.lioncore.java.model.impl.ProxyNode;
 import java.util.Collection;
 import java.util.List;
+import javax.annotation.Nonnull;
 
 public interface ClassifierInstance<T extends Classifier<T>> extends HasFeatureValues {
   /** Return all the annotations associated to this ClassifierInstance. */
+  @Nonnull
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/lionweb/lioncore/java/model/ClassifierInstance.java
```diff
@@ -1,14 +1,36 @@
...
 
+  /**
+   * Given a specific Annotation type it returns either the list of instances of that Annotation
+   * associated to the Node.
+   */
+  @Nonnull
+  List<AnnotationInstance> getAnnotations(@Nonnull Annotation annotation);
+
+  /**
+   * If an annotation instance was already associated under the Annotation link used by this
```
enikao, line 23: I thought we removed `Annotation.multiple` from M3? So we can always attach several instances of the same annotation.

ftomassetti: You are right. The logic was already like you suggested. I have now updated also the comment.

enikao:
Why not moving it to `AbstractClassifierInstance`? Annotations can have annotations.

ftomassetti:
> Why not moving it to AbstractClassifierInstance? Annotations can have annotations.

We are moving them to `ClassifierInstance,` which is more general than `AbstractClassifierInstance`. `AnnotationInstance extends `ClassifierInstance`, so the methods are also available there.

ftomassetti:
I think we covered the comments, so unless there are more doubts I would go ahead and merge this one

PR #157, opened 2024-06-28
Integrating repo client and kotlin lib

ftomassetti:
With this PR I integrate in lionweb-java the repo client and a few facilities to use LionWeb Java from Kotlin.
This is code I developed and used internally at Strumenta for a few months.

Note that both the new modules are written with Kotlin.

We could later do a significant refactoring.

I think that at this stage the most important module is the repo client. It includes functional tests that download the LW Repo and start it.


repo-client/src/main/kotlin/io/lionweb/repoclient/LionWebClient.kt
```diff
@@ -0,0 +1,734 @@
...
+            Request.Builder()
+                .url(urlBuilder.build())
+                .considerAuthenticationToken()
+                .post(body)
+                .build()
+        httpClient.newCall(request).execute().use { response ->
+            if (response.code == HttpURLConnection.HTTP_OK) {
+                val data =
+                    (response.body ?: throw IllegalStateException("Response without body when querying $url")).string()
+                var ids = rootIds.toString()
```
joswarmer, line 268: `ids`  is never used as far as I see, should these four lines be removed?

repo-client/src/main/kotlin/io/lionweb/repoclient/LionWebClient.kt
```diff
@@ -0,0 +1,734 @@
...
+                }
+            } else {
+                throw RuntimeException("Got back ${response.code}: ${response.body?.string()}")
+            }
+        }
+    }
+
+    // Nodes
+
+    fun retrieve(
```
joswarmer, line 185: Looks like this function is identical to the next retrieve function. 
Can be replaced by next function with `[rootId]` as parameter.
function body then simply is:
`return this.retrieve([rootId, withProxyParent, retrievalMode)[0]`

repo-client/src/main/kotlin/io/lionweb/repoclient/LionWebClient.kt
```diff
@@ -0,0 +1,734 @@
...
+        } else {
+            var parent = node.parent
+            if (node.parent is ProxyNode) {
+                parent = retrieve(node.parent.id!!, withProxyParent = true, retrievalMode = RetrievalMode.SINGLE_NODE)
+            }
+            return if (ancestorClass.isInstance(parent)) {
+                if (retrievalMode == RetrievalMode.SINGLE_NODE) {
+                    parent as T
+                } else {
+                    retrieve(node.parent.id!!, withProxyParent = true, retrievalMode = RetrievalMode.ENTIRE_SUBTREE) as T
```
joswarmer, line 332: Why `RetrievalModel.ENTIRE_SUBTREE`, as we are looking only at the parent?

repo-client/src/main/kotlin/io/lionweb/repoclient/LionWebClient.kt
```diff
@@ -0,0 +1,734 @@
...
+
+    fun storeTree(node: Node) {
+        treeStoringOperation(node, "store")
+    }
+
+    /**
+     * This operation is not atomic. We hope that no one is changing the parent at the very
+     * same time.
+     */
+    fun appendTree(
```
joswarmer, line 420: The lionweb-repository does return a `version` attribute in each call, that could potentially be used as a sanity check.
However, this can only be used to check whether there was any change at all to the database between calls, it is not possible (at least with the current API) to know whether a specific node has changed.

repo-client/src/main/kotlin/io/lionweb/repoclient/LionWebClient.kt
```diff
@@ -0,0 +1,734 @@
...
+        withProxyParent: Boolean = false,
+        retrievalMode: RetrievalMode = RetrievalMode.ENTIRE_SUBTREE,
+    ): Node {
+        require(rootId.isNotBlank())
+        val body: RequestBody = "{\"ids\":[\"$rootId\"] }".toRequestBody(JSON)
+        val url = "http://$hostname:$port/bulk/retrieve"
+        val urlBuilder = url.toHttpUrlOrNull()!!.newBuilder()
+        val limit =
+            when (retrievalMode) {
+                RetrievalMode.ENTIRE_SUBTREE -> "99"
```
joswarmer, line 196: Maybe use the value of the JavaScript Number.MAX_SAFE_INTEGER here? 99 is rather low.

joswarmer (approved):
I mainly looked into the LionWebClient code.
Apart from some minor remarks, it looks ok to me.

repo-client/src/main/kotlin/io/lionweb/repoclient/LionWebClient.kt
ftomassetti: Oh, wow. A lot of duplicated code. I fixed it. Thank you!

repo-client/src/main/kotlin/io/lionweb/repoclient/LionWebClient.kt
joswarmer: No problem keeping them as far as I am concerned. Maybe add a comment that this is just for debugging.

repo-client/src/main/kotlin/io/lionweb/repoclient/LionWebClient.kt
ftomassetti: Right, I created a constant called `MAX_LEVELs` that is defined as `Int.MAX_VALUE`

repo-client/src/main/kotlin/io/lionweb/repoclient/LionWebClient.kt
ftomassetti: Good idea. Done

repo-client/src/main/kotlin/io/lionweb/repoclient/LionWebClient.kt
ftomassetti: Bacause here we have identified the ancestor we are interested into, and we can now retrieve the entire subtree of the ancestor, so that it can be properly navigated. Otherwise all children would be proxies and so the user would need to do further requests

repo-client/src/main/kotlin/io/lionweb/repoclient/LionWebClient.kt
joswarmer: Just make sure it isn't bigger than the JavaScript  Number.MAX_SAFE_INTEGER , because otherwise the server might not be able to convert this into a number.

repo-client/src/main/kotlin/io/lionweb/repoclient/LionWebClient.kt
ftomassetti: Yes, we should use it, especially when we do successive calls. I addes an issue for that (https://github.com/LionWeb-io/lionweb-java/issues/159)

ftomassetti:
Thank you for the review!

repo-client/src/main/kotlin/io/lionweb/repoclient/LionWebClient.kt
joswarmer: I understand, but doesn't this mean that for each parent upwards the full subtree will be retrieved. That might result in getting quite a lot of the same nodes with each upwards parent move.

ftomassetti: No, because it does not happen for each parent but only when we identified already the right parent (see `if (ancestorClass.isInstance(parent)) {...`.

In practice we keep asking for parents in shallow mode (i.e., `retrievalMode == RetrievalMode.SINGLE_NODE`). Once we found the node of the expected type, if the user specified they want the shallow version we just return it, if instead the user specified they want the deep version we do another retrieve but this time we specify  `retrievalMode = RetrievalMode.ENTIRE_SUBTRE`. This means this particular ancestor will be retrieved twice: one in the shallow mode (`parent = retrieve(node.parent.id!!, withProxyParent = true, retrievalMode = RetrievalMode.SINGLE_NODE)` and then, when we know we have a match, again in deep mode (`retrieve(node.parent.id!!, withProxyParent = true, retrievalMode = RetrievalMode.ENTIRE_SUBTREE) as T`)

enikao:
Do we really want to mix in Kotlin with LW-Java? Then people who want to use this need to be able to build Kotlin, and contributors need to at least read Kotlin.

ftomassetti:
> Do we really want to mix in Kotlin with LW-Java? Then people who want to use this need to be able to build Kotlin, and contributors need to at least read Kotlin.

The users would have the possibility to import the modules that use Kotlin (`repo-client` and `kotlinlib`). The latter module is intended for people using Kotlin: if someone does not use Kotlin than they can just not add the module to their dependencies and safely ignore it. People developing in Java may be instead interested in using `repo-client`. It can indeed be used from Java, and if the user writes java they would not need to compile any kotlin.

Contributors would need to compile Kotlin, but they would not need to install anything manually: running gradle tasks take care of downloading everything they need. They would need to understand Kotlin if the want to contribute to the `repo-client` and `kotlinlib` modules. For the latter this is unavoidable, for the former this could be avoided rewriting such module in Java. That would require some work and make development more difficult (as Java 8 is arguably less productive than Kotlin. Maybe Java 21 is comparable to Kotlin, but adopting it means making the module unaccessible to most Java developers). Given this is a matter affecting contributors and not users, I am not sure it would make sense to convert the module to Java.

The alternative is to move these two modules into a separate project. This is doable, but it would then means that we may sometimes need to do two releases (i.e., when changing LW Java we may want to release it, go to the other project consisting of `repo-client` and `kotlinlib`, update the LW Java dependency and then do a release of that project).

I see arguments for both solutions, and if we prefer to move these two modules to a separate project it works for me.

repo-client/src/main/kotlin/io/lionweb/repoclient/LionWebClient.kt
joswarmer: You are right, I read the code incorrectly.

dslmeinte:
Is it an idea to provide a separate implementation (including a generator) for Kotlin: `lionweb-kotlin`?

I like Kotlin, but Java is still a lot bigger than Kotlin. Also: having an extra language that LionWeb is implemented probably has better optics than just `lionweb-java` which has an intermix of Java and Kotlin.

ftomassetti:
> Also: having an extra language that LionWeb is implemented probably has better optics than just lionweb-java which has an intermix of Java and Kotlin.

Good point. I will close this PR and create `lionweb-kotlin`

PR #158, opened 2024-06-29
integrate C# tests

enikao:


enikao:
@ftomassetti I tried to copy over the tests I wrote for C#. Quite a few are commented out because of API differences, and some are failing.
For now, my most pressing question concerns the generated Kotlin files. I generated them with the current Kolasu version. Besides not supporting Annotations and DataTypes, it worked ok.
But they don't implement the LionWeb Node interface. Did I take the wrong generator?

core/src/test/java/lionweb/utils/tests/CollectionAssert.java
```diff
@@ -0,0 +1,20 @@
+package lionweb.utils.tests;
+
+import org.junit.Assert;
+
+import javax.annotation.Nonnull;
+import java.util.Iterator;
+import java.util.List;
+
+public class CollectionAssert {
+    public static void AreEqual(@Nonnull List<?> expected, @Nonnull List<?> actual) {
```
dslmeinte, line 10: Not `areEqual`?

core/src/test/java/examples/shapes/dynamic/ShapesDynamic.java
```diff
@@ -0,0 +1,43 @@
...
+import io.lionweb.lioncore.java.serialization.JsonSerialization;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Entrypoint for working with the Shapes example language.
+ */
+public class ShapesDynamic {
+    public static final io.lionweb.lioncore.java.language.Language Language;
```
dslmeinte, line 29: Not `language`?

core/src/test/classes/examples/shapes/m2/BillOfMaterials.class
dslmeinte, line 1: Why are these binary `.class` files here? Are these generated directly to bytecode by Kolasu? Is documented how these files are generated?

core/src/test/java/lionweb/utils/tests/CollectionAssert.java
enikao: Yes, it's a bit less Seach+Replace atm. Want to rename it once I copied all the code.

core/src/test/java/examples/shapes/dynamic/ShapesDynamic.java
enikao: same as above, simplify Search+Replace

core/src/test/classes/examples/shapes/m2/BillOfMaterials.class
enikao: Yes, they are from Kolasu. Currently only playing around (see my open question to Federico). Once I understand how they're supposed to integrate with LW we can think of a proper way to do this.
In any case, I don't want to include any Kotlin code in this project -- would make it harder for others to consume and understand.

dslmeinte: Good points. I wouldn't mind seeing a line or 2 about this in some documentation—if even it's just saying “generated by Kolasu, instructions might come” or something.

ftomassetti:
> @ftomassetti I tried to copy over the tests I wrote for C#. Quite a few are commented out because of API differences, and some are failing.
For now, my most pressing question concerns the generated Kotlin files. I generated them with the current Kolasu version. Besides not supporting Annotations and DataTypes, it worked ok.
But they don't implement the LionWeb Node interface. Did I take the wrong generator?

I cannot find the files in the PR, but I think I understand what you mean.

Initially, I experimented with making all Kolasu nodes be _also_ lionweb nodes (the Kolasu's Node class would implement the LionWeb's Node interface). This was the solution used in the Kolasu 2.0 branch, now abandoned.

I then went for another approach in the current Kolasu 1.5 and in the Kolasu 1.6 branch under development: keeping Kolasu's Nodes separate from LionWeb Nodes but add a class that can convert between Kolasu's languages and nodes to and from LionWeb's languages and nodes. The rationale is that this gives us more flexibility, especially at a time when LionWeb's specs are still changing. 

So I think you generate Kolasu's nodes using Kolasu 1.5.* and notices they do not extend LionWeb's Nodes. 

You can:
* Use the classes generated for Kolasu and then convert them to LionWeb (see [LionWebLanguageConverter](https://github.com/Strumenta/kolasu/blob/maintenance/kolasu15/lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebLanguageConverter.kt) and [LionWebModelConverter](https://github.com/Strumenta/kolasu/blob/maintenance/kolasu15/lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt)
* We could create a new generator using the Kotlin classes for supporting LionWeb, introduced in https://github.com/LionWeb-io/lionweb-java/pull/157

Regarding the second approach, we could generate classes that look like this:
```
class FSIssue : BaseNode() {
    var message: String? by property("message")
    var severity: String? by property("severity")
    var fsPosition: FSPosition? by singleContainment("fsPosition")
    var position: MyPosition? by property("position")
}
```

Where `BaseNode` implement LionWeb's Node.

Is this what you had in mind?

enikao:
> Use the classes generated for Kolasu and then convert them to LionWeb (see [LionWebLanguageConverter](https://github.com/Strumenta/kolasu/blob/maintenance/kolasu15/lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebLanguageConverter.kt) and [LionWebModelConverter](https://github.com/Strumenta/kolasu/blob/maintenance/kolasu15/lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt)

I think that's what I did:

```
package examples.shapes.m2

import com.strumenta.kolasu.lionweb.LionWebAssociation
import com.strumenta.kolasu.model.*
import kotlin.Boolean
import kotlin.Int
import kotlin.String
import kotlin.collections.MutableList

@LionWebAssociation(key = "key-Circle")
public data class Circle(
    override var shapeDocs: Documentation,
    override var name: String,
    public var r: Int,
    public var center: Coord,
) : Shape(shapeDocs, name)

@LionWebAssociation(key = "key-Coord")
public data class Coord(
    public var x: Int,
    public var y: Int,
    public var z: Int,
) : Node()

@LionWebAssociation(key = "key-Geometry")
public data class Geometry(
    public var shapes: MutableList<IShape>,
    public var documentation: Documentation,
) : Node()

@LionWebAssociation(key = "key-IShape")
public interface IShape : NodeLike, /*manually added */ PossiblyNamed

@LionWebAssociation(key = "key-Line")
public data class Line(
    override var shapeDocs: Documentation,
    override var name: String,
    public var start: Coord,
    public var end: Coord,
) : Shape(shapeDocs, name)

@LionWebAssociation(key = "key-OffsetDuplicate")
public data class OffsetDuplicate(
    override var shapeDocs: Documentation,
    override var name: String,
    public var offset: Coord,
//    public var source: ReferenceValue<Shape>,
    public var altSource: ReferenceValue<Shape>?,
    public var docs: Documentation,
    public var secretDocs: Documentation,
) : Shape(shapeDocs, name)

@LionWebAssociation(key = "key-Shape")
public sealed class Shape(
    public open var shapeDocs: Documentation,
    open override var name: String,
) : Node(), Named, IShape

@LionWebAssociation(key = "key-CompositeShape")
public data class CompositeShape(
    override var shapeDocs: Documentation,
    override var name: String,
    public var parts: MutableList<IShape>,
    public var disabledParts: MutableList<IShape>,
    public var evilPart: IShape,
) : Shape(shapeDocs, name)

@LionWebAssociation(key = "key-ReferenceGeometry")
public data class ReferenceGeometry(
    public var shapes: ReferenceValue<IShape>?,
) : Node()

@LionWebAssociation(key = "key-Documentation")
public data class Documentation(
    public var text: String,
    public var technical: Boolean,
) : Node()

@LionWebAssociation(key = "key-BillOfMaterials")
public data class BillOfMaterials(
    public var materials: ReferenceValue<IShape>?,
    public var groups: MutableList<MaterialGroup>,
    public var altGroups: MutableList<MaterialGroup>,
    public var defaultGroup: MaterialGroup,
) : Node()

@LionWebAssociation(key = "key-MaterialGroup")
public data class MaterialGroup(
    public var matterState: MatterState,
    public var materials: ReferenceValue<IShape>,
    public var defaultShape: IShape,
) : Node()

@LionWebAssociation(key = "key-MatterState")
public enum class MatterState {
    solid,
    liquid,
    gas,
}
```
 
> We could create a new generator using the Kotlin classes for supporting LionWeb, introduced in [Integrating repo client and kotlin lib #157](https://github.com/LionWeb-io/lionweb-java/pull/157)

If we have to write a new generator anyways, I'd prefer to keep it Java-only.

core/src/test/java/examples/shapes/dynamic/ShapesDynamic.java
```diff
@@ -0,0 +1,43 @@
+// Copyright 2024 TRUMPF Laser SE and other contributors
```
ftomassetti, line 1: Should this be here?

PR #160, opened 2024-07-17
Alternative serialization formats

ftomassetti:
When measuring the performance for large insertments of nodes one of the bottle-neck that emerged is the time spent serializing, transporting, and deserializing large trees.

For this reason, I would like to introduce several optional optimizations to be used to support the processing of large trees. In this specific PR I introduce support for using, as an alternative to JSON, ProtoBuffers and FlatBuffers (both are binary formats promoted by Google, and providing libraries for multiple languages).

I also included a small program to be used for simple benchmarking. The program generate a large tree (500,000 nodes) and compare performance between the different formats. This is the output:
```
= JSON serialization (without compression) =
  serialized in 3147ms
  size 490266723 bytes
= JSON serialization (with compression) =
  serialized in 5602ms
  size 23441769 bytes
= ProtoBuf serialization =
  serialized in 1348ms
  size 31528832 bytes
= Flatbuffers serialization =
  serialized in 618ms
  size 134217728 bytes
= Comparison (protobuf against uncompressed JSON)=
  serialization time: 42.83%
  size: 6.43%
= Comparison (protobuf against compressed JSON)=
  serialization time: 24.06%
  size: 134.50%
= Comparison (flatbuffers against uncompressed JSON)=
  serialization time: 19.64%
  size: 27.38%
= Comparison (flatbuffers against compressed JSON)=
  serialization time: 11.03%
  size: 572.56%
```

In other words, serializing with flatbuffers takes 11% of the time necessary to serializing with JSON and then compressing with GZip (which is strictly needed, because otherwise the resulting size of the payload is just to large to be processed).

Note that we are measuring the time to go from the intermediate format (`SerializedChunk`) to the physical format. Going from nodes to `SerializedChunk` could also be measured and, I suppose, optimized, but I would do that as part of a separate PR.

Note: on merging, I would squash all commits in a single one


dslmeinte:
Do you intern strings? Strings take up most space once we use Protobuf-like structs. If you can replace a string with e.g. an integer (int32 would suffice, probably) index into a computed array of all strings occurring, then you can probably shave off even more.

ftomassetti:
> Do you intern strings? Strings take up most space once we use Protobuf-like structs. If you can replace a string with e.g. an integer (int32 would suffice, probably) index into a computed array of all strings occurring, then you can probably shave off even more.

Yes, I do. With protobuffers I do that explicitly: I create a stringsMap, and then I use indexes referring to that map.

Flatbuffers does that for me, as there is the method `createSharedString` that does that.

In this case we generate a tree trying to use the same strings many times, but it may not be fully representative of a real tree.

ftomassetti:
> > Do you intern strings? Strings take up most space once we use Protobuf-like structs. If you can replace a string with e.g. an integer (int32 would suffice, probably) index into a computed array of all strings occurring, then you can probably shave off even more.
> 
> Yes, I do. With protobuffers I do that explicitly: I create a stringsMap, and then I use indexes referring to that map.
> 
> Flatbuffers does that for me, as there is the method `createSharedString` that does that.
> 
> In this case we generate a tree trying to use the same strings many times, but it may not be fully representative of a real tree.

We also do that for metapointers

enikao:
Could you try to use ZStandard compression instead of GZip? It's also supported by most implementations by now, and should be a lot faster (especially on decompression).

core/src/main/java/io/lionweb/lioncore/java/serialization/AbstractSerialization.java
```diff
@@ -0,0 +1,561 @@
...
+  public void registerLanguage(Language language) {
+    getClassifierResolver().registerLanguage(language);
+    getPrimitiveValuesSerialization().registerLanguage(language);
+  }
+
+  //
+  // Serialization to chunk
+  //
+
+  public SerializedChunk serializeTreeToSerializationBlock(ClassifierInstance<?> root) {
```
enikao, line 122: Why do we have three times almost the same method? Seems like big API surface.

core/src/main/java/io/lionweb/lioncore/java/serialization/FlatBuffersSerialization.java
```diff
@@ -0,0 +1,351 @@
...
+
+  private class DeserializationHelper {
+
+    private IdentityHashMap<FBMetaPointer, MetaPointer> metaPointersCache = new IdentityHashMap<>();
+
+    public MetaPointer deserialize(FBMetaPointer classifier) {
+      if (classifier == null) {
+        throw new IllegalStateException("Classifier should not be null");
+      }
+      if (!metaPointersCache.containsKey(classifier)) {
```
enikao, line 65: Maybe use `computeIfAbsent()`? https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#computeIfAbsent-K-java.util.function.Function-

core/src/main/java/io/lionweb/lioncore/java/serialization/ProtoBufSerialization.java
```diff
@@ -0,0 +1,345 @@
...
+import io.lionweb.lioncore.java.serialization.data.MetaPointer;
+import io.lionweb.lioncore.protobuf.*;
+import java.io.*;
+import java.util.*;
+import java.util.stream.Collectors;
+
+public class ProtoBufSerialization extends AbstractSerialization {
+
+  /** This has specific support for LionCore or LionCoreBuiltins. */
+  public static ProtoBufSerialization getStandardSerialization() {
```
enikao, line 17: Maybe externalize this standard setup? It's the same for every serialization as far as I see.

enikao:
Do we also optimize this for memory consumption? Currently, the implementation has everything in memory - potentially several times: Complete JSON string, all JSON objects, all SerializedNodes, and somewhere is an API that returns a `byte[]` array of the whole contents.

Conceptually, I think there's _one_ datastructure that makes sense only when complete, and that's the tree/graph of actual nodes. All intermediate formats can be processed in a streaming fashion.

We might want to separate this aspect into another PR, though.

ftomassetti:
> Could you try to use ZStandard compression instead of GZip? It's also supported by most implementations by now, and should be a lot faster (especially on decompression).

I think it would be interesting to do this comparison, but there are two points to consider:
1. Currently we are not using ZStandard on the LionWeb Repository (as far as I know). This comparison is based on what we currently use. If we use ZStandard here, we should also ensure to support ZStandard on the LionWeb Repository
2. Looking for a Java library to support ZStandard I found 2 of them.  [aircompressor](https://github.com/airlift/aircompressor?tab=readme-ov-file), which requires a JVM v22, and then there is [zstd-jni](https://github.com/luben/zstd-jni) but I think it may require installing a native library. So both solutions do not seem a good fit. What do you think? Do you have any suggestion on the library I should use to run ZStandard compression?

core/src/main/java/io/lionweb/lioncore/java/serialization/FlatBuffersSerialization.java
ftomassetti: Good point, done in latest commit

enikao:
> What do you think? Do you have any suggestion on the library I should use to run ZStandard compression?

No, not worth the trouble. I thought it would be more common, AFAIK all web browsers support it nowadays.

core/src/main/java/io/lionweb/lioncore/java/serialization/ProtoBufSerialization.java
ftomassetti: Done: I have now introduced the class SerializationProvider and moved all those methods there, sharing the code where needed

ftomassetti:
> Do we also optimize this for memory consumption? 

No, but I think that flatbuffers should use less memory than the alternatives.
Right now I was just in a situation where the time for insertion is a major problem, and I wanted to get that solved. So far memory consumption has not been a problem, so I am ignoring it for the time being

> Currently, the implementation has everything in memory - potentially several times: Complete JSON string, all JSON objects, all SerializedNodes, and somewhere is an API that returns a `byte[]` array of the whole contents.
> 
> Conceptually, I think there's _one_ datastructure that makes sense only when complete, and that's the tree/graph of actual nodes. All intermediate formats can be processed in a streaming fashion.

Probably yes, I think we can make some optimization based on that, but I did not do that right away as I first of all wanted to check if there was any opposition to the idea of introducing more serialization format. I think we can subsequently improve the performance using this approach in a follow-up PR, if this one get approved

> 
> We might want to separate this aspect into another PR, though.

Yes, I think that would be best. Also, we could use the little benchmark I wrote to check if performance are effected and perhaps expand it to consider also memory

core/src/main/java/io/lionweb/lioncore/java/serialization/AbstractSerialization.java
enikao: I agree they are useful. The question is whether we have to put them into this API, or externalize them into helper methods -- (mostly) a matter of taste.

If they are API, we have to test all three of them for each implementation -- externalized ones only one time.

core/src/main/java/io/lionweb/lioncore/java/serialization/FlatBuffersSerialization.java
ftomassetti: (Sorry, I did not push it to the right branch, now I did)

core/src/main/java/io/lionweb/lioncore/java/serialization/ProtoBufSerialization.java
ftomassetti: (Sorry, I did not push it to the right branch, now I did)

enikao (approved):

PR #161, opened 2024-07-25
Introduce support for the BulkImport serialization to JSON, Flatbuffe…

ftomassetti:
This is needed to support the bulkImport operation, proposed in https://github.com/LionWeb-io/lionweb-repository/pull/81


ftomassetti:
> Any chance to keep this separated from the official bulk protocol?

Given these structures are useful for communication with the repo, I wanted to put them in LionWeb Kotlin. 

HowevertThe problem is that we need to introduce the BulkImport data structure for both protobuf and flatbuffers, and to do that we need to extend the schemas we already have for serializing the schema previously defined for the chunk. This is why this ended up being in this project.

Would it be enough to create a different serialization class just for the bulk import, but keeping it within this project?

enikao:
That's a start.

Do you intend to send the messages through the same channel? If not, it might work to move the new messages into a different file and import the other one, creating an independent set of source files (which might even be in another subproject).

ftomassetti:
> Do you intend to send the messages through the same channel? 

Yes, it would be used in HTTP calls to the LionWeb Repository, as for the normal inserts or retrieval

enikao:
Ok, but that would be a different http request, i.e. a new connection -- then using different messages should work? Would be different if we connected e.g. via websocket, then all messages would go through the same connection.

ftomassetti:
> Ok, but that would be a different http request, i.e. a new connection -- then using different messages should work? Would be different if we connected e.g. via websocket, then all messages would go through the same connection.

If by different messages we mean different classes, then it would mean writing some code twice, so I would avoid that if possible. The BulkImport data structure is a tiny wrapper that reuse the node definition and all of its parts (metapointers, features, etc.)

I started by creating separate Serialization classes to handle the BulkImport data structure.

enikao:
I was thinking like:

* LionWeb.core
  * lionweb.protobuf
  * lionweb.flatbuf
  * (generated classes)

* LionWeb.extension
  * lionweb-extension.protobuf
    import ../LionWeb.core/lionweb.protobuf
    ...  
  * lionweb-extension.flatbuf
    import ../LionWeb.core/lionweb.flatbuf
    ...  
  * (generated classes)


But if this is hard to do a different serializer is a good middle ground.

ftomassetti:
Ok, I think I may have found a way to use imports for both protobufs and flatbuffers and do not regenerate the imported classes. So I created the extensions module as suggested

ftomassetti:
Now that https://github.com/LionWeb-io/lionweb-repository/pull/81 has been merged, this one can be reviewed

core/src/main/java/io/lionweb/lioncore/java/serialization/ProtoBufSerialization.java
```diff
@@ -160,11 +160,13 @@ public byte[] serializeToByteArray(SerializedChunk serializedChunk) {
     return serialize(serializedChunk).toByteArray();
   }
 
-  private class SerializeHelper {
-    final Map<MetaPointer, Integer> metaPointers = new HashMap<>();
-    final Map<String, Integer> strings = new HashMap<>();
+  protected class SerializeHelper {
+    public final Map<MetaPointer, Integer> metaPointers = new HashMap<>();
```
enikao, line 164: `public` sounds suspicious for fields (same for next line)

extensions/build.gradle.kts
```diff
@@ -0,0 +1,119 @@
+import com.vanniktech.maven.publish.SonatypeHost
+import java.net.URI
+
+plugins {
+    id("java-library")
+    id("signing")
+    id("com.github.johnrengelman.shadow") version "7.1.1"
```
enikao, line 7: didn't we keep all versions in `settings.gradle` or `gradle.properties`?

extensions/build.gradle.kts
```diff
@@ -0,0 +1,119 @@
...
+    extendsFrom(configurations.testImplementation.get())
+}
+
+dependencies {
+    implementation(project(":core"))
+
+    // Use JUnit test framework.
+    testImplementation(libs.junit)
+
+    implementation("com.google.protobuf:protobuf-java:4.27.2")
```
enikao, line 30: another set of version numbers

extensions/build.gradle.kts
```diff
@@ -0,0 +1,119 @@
...
+
+tasks.jacocoTestReport {
+    dependsOn(tasks.test) // tests are required to run before generating the report
+}
+
+protobuf {
+    protoc {
+        protoc {
+            // Apple Silicon processor would look for an unexisting platform-specific variant
+            artifact = "com.google.protobuf:protoc:4.27.2" + if (osdetector.os == "osx") ":osx-x86_64" else ""
```
enikao, line 107: another version

extensions/build.gradle.kts
```diff
@@ -0,0 +1,119 @@
...
+    sourceCompatibility = JavaVersion.toVersion(jvmVersion)
+    targetCompatibility = JavaVersion.toVersion(jvmVersion)
+}
+
+tasks.jacocoTestReport {
+    dependsOn(tasks.test) // tests are required to run before generating the report
+}
+
+protobuf {
+    protoc {
```
enikao, line 104: really two nested `protoc` blocks?

extensions/src/main/java/io/lionweb/serialization/flatbuffers/FBAttachPoint.java
```diff
@@ -0,0 +1,116 @@
+// automatically generated by the FlatBuffers compiler, do not modify
```
enikao, line 1: maybe move generated classes into a `.gen.` subpackage?

core/src/main/java/io/lionweb/lioncore/java/serialization/ProtoBufSerialization.java
ftomassetti: Changed

extensions/build.gradle.kts
ftomassetti: Removed

extensions/build.gradle.kts
ftomassetti: I added a versions catalog and put it there

extensions/build.gradle.kts
ftomassetti: I added a versions catalog and put it there

extensions/build.gradle.kts
ftomassetti: I added a versions catalog and put it there

extensions/src/main/java/io/lionweb/serialization/extensions/ExtraFlatBuffersSerialization.java
```diff
@@ -0,0 +1,184 @@
...
+import io.lionweb.serialization.flatbuffers.gen.*;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+
+/** It contains the logic to serialize non-standard messages. */
+public class ExtraFlatBuffersSerialization extends FlatBuffersSerialization {
+
+  public byte[] serializeBulkImport(BulkImport bulkImport) {
```
enikao, line 23: This is a pretty long method, consider splitting it up.

settings.gradle
```diff
@@ -1,8 +1,13 @@
+pluginManagement {
+    plugins {
+        id 'org.jetbrains.kotlin.jvm' version '2.0.0'
```
enikao, line 3: Why do we use kotlin in lw-java?

ftomassetti: At some point I must have added a Kotlin file by mistake and IDEA must have added this automatically. I removed it now

extensions/src/main/java/io/lionweb/serialization/extensions/ExtraFlatBuffersSerialization.java
ftomassetti: I extracted two methods from this one

ftomassetti:
Thank you @enikao !

Issue #165, opened 2024-10-01
Implement workaround for EAttributes with high multiplicity on the M1 level


Ulyana-F1re:
In LionWeb properties can be 0..1 or 1..1, but they can not have higher multiplicity (so 0..* or 1..* are not supported). In EMF ecore EAttributes can have high multiplicity. To support import of ecore metamodels with EAttributes of higher multiplicity, I updated `EMFMetamodelImporter` as following: we introduce an intermediate `Concept` that holds the property corresponding to the `EAttribute`, and this concept is added to the parent concept as a containment with high multiplicity.

Now we need to update `EMFModelImporter` and `EMFModelExporter` so that they are in synch with this meta-model transformation, so that we have a bi-diractional mapping between the original `EAtribute` and the introduced auxiliary LionCore containment on the M1 level.

ftomassetti:
Could it make sense to add an annotation on these "synthetic" concepts, so that we know they need special treatment when importing or exporting models?

enikao:
I had to solve a similar problem with MPS references (`0..1`) vs. LW references (`0..*`), with the same approach. I also have annotations like @ftomassetti proposes, they are very helpful (especially for multiple import/export scenarios).

Ulyana-F1re:
To me it looks natural to implement such model transformations (also the one from [#166]) in the ecore world, using for example Xtend. That's what I also do in Rascal: all model transformations between LionCore and Rascal, on the M1 and M2, are implemented in Rascal. Does it make sense to use Xtend here too?

enikao:
I wouldn't use Xtend for anything but M2T transformations. For M2M, the only real benefit it brings over modern Java is memoization. But we'd introduce a dependency on a kind of "fringe" technology. Doesn't sound like it's worth the risk to me.

dslmeinte:
And memoization is easy enough to implement in a GPL (even in a transparent way using HOFs) as well.

PR #166, opened 2024-10-01
Workaround for EAttribute with high multiplicity, on the M2 level

Ulyana-F1re:
In LionWeb properties can be 0..1 or 1..1, but they can not have higher multiplicity (so 0..* or 1..* are not supported). In EMF ecore `EAttributes` can have high multiplicity. To support import of ecore metamodels with `EAttributes` of higher multiplicity, I updated `EMFMetamodelImporte`r as following: we introduce an intermediate LionCore `Concept` that holds the `Property` that corresponds to the `EAttribute`, and this concept is added to the parent concept as a `Containment` with high multiplicity.

As a follow-up we need to implement such a mapping on the M1 level, see issue [#165 ]

### DIFF
--- a/emf/src/main/java/io/lionweb/lioncore/java/emf/EMFMetamodelImporter.java
+++ b/emf/src/main/java/io/lionweb/lioncore/java/emf/EMFMetamodelImporter.java
@@ -8,6 +8,8 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Objects;
 import javax.annotation.Nonnull;
+
+import io.lionweb.lioncore.java.model.impl.DynamicAnnotationInstance;
 import org.eclipse.emf.ecore.*;
 import org.eclipse.emf.ecore.resource.Resource;
 
@@ -43,8 +45,7 @@ public class EMFMetamodelImporter extends AbstractEMFImporter<Language> {
   public Language importEPackage(EPackage ePackage) {
     Language metamodel = new Language(ePackage.getName());
     metamodel.setVersion("1");
-    metamodel.setID(ePackage.getName());
-    metamodel.setKey(ePackage.getName());
+    setIDAndKey(metamodel, ePackage.getName());
 
     // Initially we just create empty concepts, later we populate the features as they could refer
     // to EClasses which we meet later on in the EPackage
@@ -53,14 +54,12 @@ public class EMFMetamodelImporter extends AbstractEMFImporter<Language> {
         EClass eClass = (EClass) eClassifier;
         if (eClass.isInterface()) {
           Interface iface = new Interface(metamodel, eClass.getName());
-          iface.setID(ePackage.getName() + "-" + iface.getName());
-          iface.setKey(ePackage.getName() + "-" + iface.getName());
+          setIDAndKey(iface, ePackage.getName() + "-" + iface.getName());
           metamodel.addElement(iface);
           conceptsToEClassesMapping.registerMapping(iface, eClass);
         } else {
           Concept concept = new Concept(metamodel, eClass.getName());
-          concept.setID(ePackage.getName() + "-" + concept.getName());
-          concept.setKey(ePackage.getName() + "-" + concept.getName());
+          setIDAndKey(concept, ePackage.getName() + "-" + concept.getName());
           concept.setAbstract(false);
           metamodel.addElement(concept);
           conceptsToEClassesMapping.registerMapping(concept, eClass);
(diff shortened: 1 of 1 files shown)


Ulyana-F1re:
It would be nice to log such changes to the metamodel, but I don't see whether any logger is used in the rest of the project.

emf/src/main/java/io/lionweb/lioncore/java/emf/EMFMetamodelImporter.java
```diff
@@ -146,17 +146,45 @@ private void processStructuralFeatures(
...
+          property.setID(ePackage.getName() + "-" + holderConcept.getName() + "-" + property.getName());
+          property.setKey(ePackage.getName() + "-" + holderConcept.getName() + "-" + property.getName());
+          holderConcept.addFeature(property);
+          property.setOptional(false);
+          DataType<DataType> propertyType =
+                  dataTypeMapping.convertEClassifierToDataType(eFeature.getEType());
+          Objects.requireNonNull(propertyType, "Cannot convert type " + eFeature.getEType());
+          property.setType(propertyType);
+
+          Containment containment = new Containment(eFeature.getName(), classifier);
```
dslmeinte, line 181: I see multiple occurrences of creating an M2 element and then setting its ID and key in very predictable ways: could you make a helper function for this? It would help with the readability of this code.

emf/src/main/java/io/lionweb/lioncore/java/emf/EMFMetamodelImporter.java
```diff
@@ -146,17 +146,45 @@ private void processStructuralFeatures(
...
-        property.setKey(ePackage.getName() + "-" + classifier.getName() + "-" + eFeature.getName());
-        classifier.addFeature(property);
-        property.setOptional(!eAttribute.isRequired());
-        DataType<DataType> propertyType =
-            dataTypeMapping.convertEClassifierToDataType(eFeature.getEType());
-        Objects.requireNonNull(propertyType, "Cannot convert type " + eFeature.getEType());
-        property.setType(propertyType);
-        if (eAttribute.isMany()) {
-          throw new IllegalArgumentException("EAttributes with upper bound > 1 are not supported");
+        // EAttributes with upper bound > 1 are not supported in LionWeb
```
dslmeinte, line 149: I think this comment doesn't need to be there: you're saying the same thing on line 161 already, and you first handle the upper bound <= 1 case.

ftomassetti:
> It would be nice to log such changes to the metamodel, but I don't see whether any logger is used in the rest of the project.

Maybe we can use annotations for that?

Ulyana-F1re:
> > It would be nice to log such changes to the metamodel, but I don't see whether any logger is used in the rest of the project.
> 
> Maybe we can use annotations for that?

Annotations is a nice idea! I can add them.
I still would log this information too: it provides an overview of what happened during the transformation (i.e. how many concepts in my metamodel were not conformant with the LionWeb format). While with the annotation I would need to look through the language concepts to find them.

emf/src/main/java/io/lionweb/lioncore/java/emf/EMFMetamodelImporter.java
Ulyana-F1re: Shall we not use constructors for this? I see that classes like `Concept`, `Containment`, `Interface` have constructors with `key` and `id` as their parameters, but `Property`, `Enumeration` and `EnumerationLiteral` - not.
Or do you mean a helper function for constructing IDs and Keys?

dslmeinte: I originally meant helper functions, but it makes more sense to align the styles of constructors.

ftomassetti:
I think this PR is close to be ready. Could we revise the comments @dslmeinte added and get this one over the line?

Ulyana-F1re:
I have implemented an ecore-to-ecore transformation in QVTo, that takes care of making an ecore metamodel compatible with LionWeb format. I like this approach more, as it is more transparent and I can support it on M1 level with a corresponding QVTo transformation too.

dslmeinte (approved):

ftomassetti:
Thank you @Ulyana-F1re !
I see @dslmeinte has approved it, so I will go on and merge it.

dslmeinte:
> I see @dslmeinte has approved it, so I will go on and merge it.

More dangerous words have rarely been spoken... 😃

PR #167, opened 2024-11-18
Support multiple LionWeb versions

ftomassetti:
The goal is to support, with the same codebase, all the LionWeb versions, provided they are not mixed together.

Many APIs now get an optional parameter indicating the LionWeb version to be used. If no value is provided the current one is used.

Not also that all "meta" nodes (Concept, Enumeration, Feature, etc.) needs to know to which LionWeb version they refer to.

Probably more checks should be put in place to identify cases in which multiple versions are mixed.

Regarding serialization tests, in all the cases in which we had files, we kept them untouched, so we check the serialization works for v2023.1. At some point we may duplicate some of those tests to check they still work with v2024.1 (with the appropriate, minimal, adaptations).

### DIFF
--- a/core/src/main/java/io/lionweb/lioncore/java/LionWebVersion.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/LionWebVersion.java
@@ -11,16 +11,16 @@ public enum LionWebVersion {
   v2023_1("2023.1"),
   v2024_1("2024.1");
   public static final LionWebVersion currentVersion = LionWebVersion.v2024_1;
-  private @Nonnull String value;
+  private @Nonnull String versionString;
 
-  LionWebVersion(@Nonnull String value) {
-    Objects.requireNonNull(value, "Value should not be null");
-    this.value = value;
+  LionWebVersion(@Nonnull String versionString) {
+    Objects.requireNonNull(versionString, "versionString should not be null");
+    this.versionString = versionString;
   }
 
   public static LionWebVersion fromValue(String serializationFormatVersion) {
     for (LionWebVersion lionWebVersion : values()) {
-      if (lionWebVersion.getValue().equals(serializationFormatVersion)) {
+      if (lionWebVersion.getVersionString().equals(serializationFormatVersion)) {
         return lionWebVersion;
       }
     }
@@ -28,7 +28,7 @@ public enum LionWebVersion {
         "Invalid serializationFormatVersion: " + serializationFormatVersion);
   }
 
-  public @Nonnull String getValue() {
-    return value;
+  public @Nonnull String getVersionString() {
+    return versionString;
   }
 }
--- a/core/src/main/java/io/lionweb/lioncore/java/LionWebVersion.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/LionWebVersion.java
@@ -1,12 +1,20 @@
 package io.lionweb.lioncore.java;
 
+import java.util.Objects;
+import javax.annotation.Nonnull;
+
+/**
+ * A LionWeb Version. Note that the version is used to refer to the specifications but also to the
+ * versions of LionCore and LionCore Builtins, as they should always be aligned.
+ */
 public enum LionWebVersion {
   v2023_1("2023.1"),
   v2024_1("2024.1");
   public static final LionWebVersion currentVersion = LionWebVersion.v2024_1;
-  private String value;
+  private @Nonnull String value;
 
-  LionWebVersion(String value) {
+  LionWebVersion(@Nonnull String value) {
+    Objects.requireNonNull(value, "Value should not be null");
     this.value = value;
   }
 
@@ -20,7 +28,7 @@ public enum LionWebVersion {
         "Invalid serializationFormatVersion: " + serializationFormatVersion);
   }
 
-  public String getValue() {
+  public @Nonnull String getValue() {
     return value;
   }
 }
--- a/core/src/main/java/io/lionweb/lioncore/java/LionWebVersion.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/LionWebVersion.java
@@ -16,7 +16,8 @@ public enum LionWebVersion {
         return lionWebVersion;
       }
     }
-    throw new IllegalArgumentException("Invalid serializationFormatVersion: " + serializationFormatVersion);
+    throw new IllegalArgumentException(
+        "Invalid serializationFormatVersion: " + serializationFormatVersion);
   }
 
   public String getValue() {
--- a/core/src/main/java/io/lionweb/lioncore/java/LionWebVersion.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/LionWebVersion.java
@@ -1,0 +1,7 @@
+package io.lionweb.lioncore.java;
+
+public enum LionWebVersion {
+  v2023_1,
+  v2024_1;
+  public static final LionWebVersion currentVersion = LionWebVersion.v2024_1;
+}
--- a/core/src/main/java/io/lionweb/lioncore/java/LionWebVersion.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/LionWebVersion.java
@@ -1,7 +1,25 @@
 package io.lionweb.lioncore.java;
 
 public enum LionWebVersion {
-  v2023_1,
-  v2024_1;
+  v2023_1("2023.1"),
+  v2024_1("2024.1");
   public static final LionWebVersion currentVersion = LionWebVersion.v2024_1;
+  private String value;
+
+  LionWebVersion(String value) {
+    this.value = value;
+  }
+
+  public static LionWebVersion fromValue(String serializationFormatVersion) {
+    for (LionWebVersion lionWebVersion : values()) {
+      if (lionWebVersion.getValue().equals(serializationFormatVersion)) {
+        return lionWebVersion;
+      }
+    }
+    throw new IllegalArgumentException("Invalid serializationFormatVersion: " + serializationFormatVersion);
+  }
+
+  public String getValue() {
+    return value;
+  }
 }
--- a/core/src/main/java/io/lionweb/lioncore/java/language/Enumeration.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/language/Enumeration.java
@@ -36,6 +36,6 @@ public class Enumeration extends DataType<Enumeration> implements NamespaceProvi
 
   @Override
   public Concept getClassifier() {
-    return LionCore.getEnumeration();
+    return LionCore.getEnumeration(getLionWebVersion());
   }
 }
--- a/core/src/main/java/io/lionweb/lioncore/java/language/Enumeration.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/language/Enumeration.java
@@ -1,5 +1,6 @@
 package io.lionweb.lioncore.java.language;
 
+import io.lionweb.lioncore.java.LionWebVersion;
 import io.lionweb.lioncore.java.self.LionCore;
 import java.util.List;
 import java.util.Objects;
@@ -7,6 +8,10 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 public class Enumeration extends DataType<Enumeration> implements NamespaceProvider {
+  public Enumeration(@Nonnull LionWebVersion lionWebVersion) {
+    super(lionWebVersion);
+  }
+
   public Enumeration() {
     super();
   }
--- a/core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
@@ -1,11 +1,9 @@
 package io.lionweb.lioncore.java.language;
 
 import io.lionweb.lioncore.java.LionWebVersion;
-import io.lionweb.lioncore.java.serialization.JsonSerialization;
-
-import javax.annotation.Nonnull;
 import java.util.HashMap;
 import java.util.Map;
+import javax.annotation.Nonnull;
 
 public class LionCoreBuiltins extends Language {
   private static final Map<LionWebVersion, LionCoreBuiltins> INSTANCES = new HashMap<>();
@@ -26,7 +24,8 @@ public class LionCoreBuiltins extends Language {
     Concept node = new Concept(lionWebVersion, this, "Node").setID("LionCore-builtins-Node");
     node.setAbstract(true);
 
-    Interface iNamed = new Interface(lionWebVersion, this, "INamed").setID("LionCore-builtins-INamed");
+    Interface iNamed =
+        new Interface(lionWebVersion, this, "INamed").setID("LionCore-builtins-INamed");
     iNamed.addFeature(
         Property.createRequired(lionWebVersion, "name", string)
             .setID("LionCore-builtins-INamed-name")
--- a/core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
@@ -16,7 +16,7 @@ public class LionCoreBuiltins extends Language {
     super(lionWebVersion, "LionCore_builtins");
     setID("LionCore-builtins");
     setKey("LionCore-builtins");
-    setVersion(lionWebVersion.getValue());
+    setVersion(lionWebVersion.getVersionString());
     PrimitiveType string = new PrimitiveType(lionWebVersion, this, "String");
     new PrimitiveType(lionWebVersion, this, "Boolean");
     new PrimitiveType(lionWebVersion, this, "Integer");
@@ -50,26 +50,6 @@ public class LionCoreBuiltins extends Language {
     return getInstance(LionWebVersion.currentVersion);
   }
 
-  public static PrimitiveType getString() {
-    return getInstance().getPrimitiveTypeByName("String");
-  }
-
-  public static PrimitiveType getInteger() {
-    return getInstance().getPrimitiveTypeByName("Integer");
-  }
-
-  public static PrimitiveType getBoolean() {
-    return getInstance().getPrimitiveTypeByName("Boolean");
-  }
-
-  public static Interface getINamed() {
-    return getInstance().getInterfaceByName("INamed");
-  }
-
-  public static Concept getNode() {
-    return getInstance().getConceptByName("Node");
-  }
-
   public static LionCoreBuiltins getInstance(@Nonnull LionWebVersion lionWebVersion) {
     Objects.requireNonNull(lionWebVersion, "lionWebVersion should not be null");
     if (!INSTANCES.containsKey(lionWebVersion)) {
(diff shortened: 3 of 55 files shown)


core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
```diff
@@ -1,27 +1,36 @@
...
+  // We may have one instance of this class per LionWeb version, and we initialize
+  // them lazily
+  private static final Map<LionWebVersion, LionCoreBuiltins> INSTANCES = new HashMap<>();
 
   /** This is private to prevent instantiation and enforce the Singleton pattern. */
-  private LionCoreBuiltins() {
-    super("LionCore_builtins");
+  private LionCoreBuiltins(@Nonnull LionWebVersion lionWebVersion) {
+    super(lionWebVersion, "LionCore_builtins");
     setID("LionCore-builtins");
```
enikao, line 17: This means we have more than one node with the same id. Do we want this? It violates one of our fundamental assumptions -- every node has a unique id.

We also have an (unresolved) issue on this: https://github.com/LionWeb-io/specification/issues/279

core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
```diff
@@ -1,27 +1,36 @@
...
-    setVersion(JsonSerialization.DEFAULT_SERIALIZATION_FORMAT);
-    PrimitiveType string = new PrimitiveType(this, "String");
-    new PrimitiveType(this, "Boolean");
-    new PrimitiveType(this, "Integer");
-    new PrimitiveType(this, "JSON");
+    setVersion(lionWebVersion.getValue());
+    PrimitiveType string = new PrimitiveType(lionWebVersion, this, "String");
+    new PrimitiveType(lionWebVersion, this, "Boolean");
+    new PrimitiveType(lionWebVersion, this, "Integer");
+    if (lionWebVersion.equals(LionWebVersion.v2023_1)) {
```
enikao, line 23: This kind of compatibility will be pretty hard to maintain on the long term.

core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
```diff
@@ -38,30 +47,61 @@ private LionCoreBuiltins() {
...
-    return INSTANCE.getInterfaceByName("INamed");
+    return getInstance().getInterfaceByName("INamed");
   }
 
   public static Concept getNode() {
-    return INSTANCE.getConceptByName("Node");
+    return getInstance().getConceptByName("Node");
+  }
+
+  public static LionCoreBuiltins getInstance(@Nonnull LionWebVersion lionWebVersion) {
```
enikao, line 73: I think it will be close to impossible to control that all users of this method will use it with the correct version, so mixing them will be pretty common.

core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
```diff
@@ -38,30 +47,61 @@ private LionCoreBuiltins() {
   }
 
   public static LionCoreBuiltins getInstance() {
```
enikao, line 49: For me it's more intuitive to have several overloads of the same method next to each other.

core/src/main/java/io/lionweb/lioncore/java/language/Property.java
```diff
@@ -57,11 +89,28 @@ public Property() {
     super();
   }
 
+  public Property(
+      @Nonnull LionWebVersion lionWebVersion,
+      @Nullable String name,
+      @Nullable Classifier container,
+      @Nonnull String id) {
+    // TODO verify that the container is also a NamespaceProvider
```
enikao, line 97: Do we still have this concept?

core/src/main/java/io/lionweb/lioncore/java/language/Property.java
```diff
@@ -57,11 +89,28 @@ public Property() {
...
+    super(lionWebVersion, name, container, id);
+  }
+
   public Property(@Nullable String name, @Nullable Classifier container, @Nonnull String id) {
     // TODO verify that the container is also a NamespaceProvider
```
enikao, line 102: Do we still have this concept?

core/src/main/java/io/lionweb/lioncore/java/language/Property.java
```diff
@@ -57,11 +89,28 @@ public Property() {
...
     super(name, container, id);
   }
 
+  public Property(
+      @Nonnull LionWebVersion lionWebVersion,
+      @Nullable String name,
+      @Nullable Classifier container) {
+    // TODO verify that the container is also a NamespaceProvider
```
enikao, line 110: Do we still have this concept?

core/src/main/java/io/lionweb/lioncore/java/language/Property.java
```diff
@@ -57,11 +89,28 @@ public Property() {
...
+    super(lionWebVersion, name, container);
+  }
+
   public Property(@Nullable String name, @Nullable Classifier container) {
     // TODO verify that the container is also a NamespaceProvider
```
enikao, line 115: Do we still have this concept?

core/src/main/java/io/lionweb/lioncore/java/model/impl/M3Node.java
```diff
@@ -14,15 +15,19 @@
...
  * possible.
+ *
+ * <p>Each M3Node is connected to a specific version of lionWebVersion, as these elements may behave
+ * differently depending on the version of LionWeb they are representing.
  */
 public abstract class M3Node<T extends M3Node> extends AbstractClassifierInstance<Concept>
     implements Node {
-  private String id;
-  private Node parent;
+  private @Nonnull LionWebVersion lionWebVersion;
```
enikao, line 28: maybe `final`?

core/src/main/java/io/lionweb/lioncore/java/serialization/AbstractSerialization.java
```diff
@@ -22,7 +23,6 @@
  * and the actual physical formats is done in other classes.
  */
 public abstract class AbstractSerialization {
-  public static final String DEFAULT_SERIALIZATION_FORMAT = "2023.1";
```
enikao, line 25: Do we care about compatibility with existing implementations? This breaks existing users.

core/src/main/java/io/lionweb/lioncore/java/serialization/AbstractSerialization.java
```diff
@@ -48,7 +48,15 @@ public abstract class AbstractSerialization {
   protected UnavailableNodePolicy unavailableReferenceTargetPolicy =
       UnavailableNodePolicy.THROW_ERROR;
 
+  private @Nonnull LionWebVersion lionWebVersion;
```
enikao, line 51: maybe `final`?

core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
enikao, line 49: What about declaring this method `@Deprecated`? It's almost certainly wrong to use it.

Same for all the other version-less methods.

core/src/main/java/io/lionweb/lioncore/java/self/LionCore.java
```diff
@@ -1,118 +1,196 @@
...
+
+  public static @Nonnull Concept getProperty(@Nonnull LionWebVersion lionWebVersion) {
+    return getInstance(lionWebVersion).requireConceptByName("Property");
+  }
+
+  public static @Nonnull Concept getReference(@Nonnull LionWebVersion lionWebVersion) {
+    return getInstance(lionWebVersion).requireConceptByName("Reference");
+  }
+
+  public static @Nonnull Language getInstance() {
```
enikao, line 139: What about declaring this method `@Deprecated`? It's almost certainly wrong to use it.

Same for all the other version-less methods.

core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
ftomassetti: I thought that the specs dictated these IDs, and I think this should be the case. Otherwise when we load models from different sources in the same repository we may have confusion as we refer to nodes everyone is aware of (e.g., the node for `Concept`) through different IDs.

For now however clashing ids should not be a problem because of the decision of considering different versions of lionweb incompatible, so we cannot mix together in the same "universe" nodes from different versions.

core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
ftomassetti: I am not sure this would be the case as I am not sure how many versions we will have and how much they will vary from each other. For now, it seems the simplest approach. Or do you think we should refactor it?

core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
ftomassetti: Yes, that could be the case. However I would expect these methods to be used in "infrastructure" code (like the repository, libraries to implement migrations, etc.). Everyday users would probably not use these methods frequently. Also, I think that most users would just use the variants not specifying a version because they may want to always use the latest one, unless they specifically need to deal with "legacy" models

enikao: I'm currently trying in the C# version whether I can use the compiler to guarantee this (would also dictate the [issue above](https://github.com/LionWeb-io/lionweb-java/pull/167#discussion_r1847186860)). I'll let you know how I fare.

core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
ftomassetti: Right, I now grouped overloaded methods

core/src/main/java/io/lionweb/lioncore/java/language/Property.java
ftomassetti: Yes

core/src/main/java/io/lionweb/lioncore/java/language/Property.java
ftomassetti: Yes

core/src/main/java/io/lionweb/lioncore/java/language/Property.java
ftomassetti: Yes

core/src/main/java/io/lionweb/lioncore/java/language/Property.java
ftomassetti: Yes

core/src/main/java/io/lionweb/lioncore/java/model/impl/M3Node.java
ftomassetti: Right, it is now final

core/src/main/java/io/lionweb/lioncore/java/serialization/AbstractSerialization.java
ftomassetti: I have restored it and marked as deprecated

core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
ftomassetti: I think most users would use the variants that are version-less, because they want to use the latest version and when they migrate to a new version this would be seemless. I think very few users will have a reason to use a specific version and not just the latest one

ftomassetti: I have also done that in LionCore

core/src/main/java/io/lionweb/lioncore/java/LionWebVersion.java
```diff
@@ -0,0 +1,34 @@
...
+      if (lionWebVersion.getValue().equals(serializationFormatVersion)) {
+        return lionWebVersion;
+      }
+    }
+    throw new IllegalArgumentException(
+        "Invalid serializationFormatVersion: " + serializationFormatVersion);
+  }
+
+  public @Nonnull String getValue() {
+    return value;
```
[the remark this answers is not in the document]
ftomassetti: Done!

core/src/main/java/io/lionweb/lioncore/java/serialization/AbstractSerialization.java
ftomassetti: Done

core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
enikao: I was not successful in C#. However, my guess is it would work in Java. It essentially failed because I cannot write `SomeType<?>` in C# -- but Java allows it.

enikao: This was my idea:

I have a marker interface `IVersionDependent` that's implemented by every class that has to work differently for different versions. It has a generic type parameter `T extends IVersion` to bind it to some version.
Each version is represented by a class that implements `IVersion`.

Our guinea pig is a `Serializer`. It has to process builtin primitives differently depending on the version.
Thus we define an interface `IProcessBuiltinPrimitives`, and one implementation for each version.

When we create a serializer for a specific version, we have to create the correct variant for processing builtin primitives.
This is the part that doesn't work in C#, but should work in Java.

Not shown here: All M2 languages also implement `IVersionDependent`, so the compiler makes sure we only pass languages to the serializer with the correct version: `Serializer<TSerializerVersion> { void RegisterLanguage<TLangVersion>(Language<TLangVersion> language) where TLangVersion : TSerializerVersion {} }`

```csharp

interface IVersion;

class V1 : IVersion;
class V2 : IVersion;

interface IVersionDependent<T> where T : IVersion;

class Serializer<T> : IVersionDependent<T> where T : IVersion
{
    public string Name { get; } = "me!";
    public IDelegate<T> Deleg { get; init; }

    public void Run() => Deleg.DoIt(this);
}

interface IProcessBuiltinPrimitives<T> where T : IVersion
{
    public void DoIt(Serializer<T> serial);
}

abstract class ProcessBuiltinPrimitives<T> : IProcessBuiltinPrimitives<T> where T : IVersion
{
    public abstract void DoIt(Serializer<T> serial);
}

class ProcessBuiltinPrimitives_V1 : ProcessBuiltinPrimitives<V1>
{
    public override void DoIt(Serializer<T> serial) => Console.WriteLine($"V1{serial.Name}");
}

class ProcessBuiltinPrimitives_V2: ProcessBuiltinPrimitives<V2>
{
    public override void DoIt(Serializer<T> serial) => Console.WriteLine($"V2{serial.Name}");
}

class SerializerBuilder<Tx> : IVersionDependent<Tx> where Tx : IVersion
{
    public Serializer<T> CreateSerializer<T>(T version) where T : Tx
    {
        // Would need IProcessBuiltinPrimitives<?>
        IProcessBuiltinPrimitives<T> deleg = version switch
        {
            V1 => new ProcessBuiltinPrimitives_V1(),
            V2 => new ProcessBuiltinPrimitives_V2()
        };

        var serial = new Serializer<T>() { Deleg = deleg };
        return serial;
    }
}
```

enikao: It does work:

```java
public class Bla {
    interface IVersion{}

    static class V1 implements IVersion{
        public String toString() { return "V1"; }
    }
    static class V2 implements IVersion{
        public String toString() { return "V2"; }
    }

    interface IVersionDependent<T extends IVersion> {}

    static class Language<T extends IVersion> implements IVersionDependent<T> {
        private final T version;
        private final String name;

        public Language(T version, String name) {
            this.version = version;
            this.name = name;
        }

        public String toString() { return "Language{" + "version=" + version + ", name='" + name + '\'' + '}'; }
    }

    static class Serializer<T extends IVersion> implements IVersionDependent<T>
    {
        private final IProcessBuiltinPrimitives<T> deleg;
        public Serializer(IProcessBuiltinPrimitives<T> deleg) { this.deleg = deleg; }

        public String getName() { return "me"; }
        public String Run(Language<T> language) {return deleg.DoIt(this) + " " + language;}
    }

    interface IProcessBuiltinPrimitives<T extends IVersion>
    {
        String DoIt(Serializer<T> serial);
    }

    static class ProcessBuiltinPrimitives_V1 implements IProcessBuiltinPrimitives<V1>
    {
        public String DoIt(Serializer<V1> serial) { return "V1" + serial.getName(); }
    }

    static class ProcessBuiltinPrimitives_V2 implements IProcessBuiltinPrimitives<V2>
    {
        public String DoIt(Serializer<V2> serial) { return "V2" + serial.getName(); }
    }

    static class SerializerBuilder<T extends IVersion> implements IVersionDependent<T>
    {
        private final T version;

        public SerializerBuilder(T version) { this.version = version; }

        public Serializer<T> CreateSerializer()
        {
            IProcessBuiltinPrimitives<T> deleg = (IProcessBuiltinPrimitives<T>) (version instanceof V1 ?  new ProcessBuiltinPrimitives_V1() :  new ProcessBuiltinPrimitives_V2());

            return new Serializer<>(deleg);
        }
    }

    @Test
    public void test() {
        Language<V1> v1Lang = new Language<>(new V1(), "v1Lang");
        Language<V2> v2Lang = new Language<>(new V2(), "v2Lang");

        Assert.assertEquals("V1me Language{version=V1, name='v1Lang'}", new SerializerBuilder<>(new V1()).CreateSerializer().Run(v1Lang));
        // Doesn't compile
        // Assert.assertEquals("V2me Language{version=V1, name='v1Lang'}", new SerializerBuilder<V2>(new V2()).CreateSerializer().Run(v1Lang));
        Assert.assertEquals("V2me Language{version=V2, name='v2Lang'}", new SerializerBuilder<>(new V2()).CreateSerializer().Run(v2Lang));
    }
}
```

core/src/main/java/io/lionweb/lioncore/java/language/Enumeration.java
```diff
@@ -1,12 +1,17 @@
 package io.lionweb.lioncore.java.language;
 
+import io.lionweb.lioncore.java.LionWebVersion;
 import io.lionweb.lioncore.java.self.LionCore;
 import java.util.List;
 import java.util.Objects;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 public class Enumeration extends DataType<Enumeration> implements NamespaceProvider {
```
dslmeinte, line 10: `NamespaceProvider` is not an entity in the M3 anymore. It might be convenient here, but it's also confusing. Also: can we get rid of TODOs that mention it?

core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
ftomassetti: I thought about using generics, but I was afraid it would make the code more cumbersome to use. Forcing users to write more complex code and having casts here and there.

Admittedly I did not think about using a token type for versions as you did.

Perhaps we can do an experiment with this idea to see how complex the code gets and then decide if the safety it provides is worth the complexity. Would this make sense?

core/src/main/java/io/lionweb/lioncore/java/language/Enumeration.java
ftomassetti: It turns out that the TODO messages are outdated as all the containers we use implement NamespaceProvider, so I removed all such comments

core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
ftomassetti: I started the experiment. I am writing a lot of `<` and `>` :D 
I will share it as soon as I can

dslmeinte:
> After experimenting with using generics in #170 I think that the resulting code is cumbersome to write and to use, so I would prefer going for this PR, where versions compatibility is checked at runtime and not a compile time. @enikao @dslmeinte what do you think?

Sounds like a good decision. If, later on, we do find a way to achieve version-safety through generics (or a similar strategy) it should be doable to retrofit that onto these changes.

core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
ftomassetti: This will be handled in https://github.com/LionWeb-io/lionweb-java/issues/172

PR #170, opened 2024-11-22
Experiment about using generics to catch version mismatches

ftomassetti:
This is an incomplete experiment to check what it means to use generics to catch version mismatches, using a "version token type", as proposed by @enikao .

To make this work, more work should be put in.

Most classes will need to be touched and I am not 100% sure we can make the generics always work.

### DIFF
--- a/core/src/integrationTest/java/io/lionweb/lioncore/java/testset/ALanguageTestset.java
+++ b/core/src/integrationTest/java/io/lionweb/lioncore/java/testset/ALanguageTestset.java
@@ -1,6 +1,6 @@
 package io.lionweb.lioncore.java.testset;
 
-import io.lionweb.lioncore.java.LionWebVersion;
+import io.lionweb.lioncore.java.versions.LionWebVersion;
 import io.lionweb.lioncore.java.language.Language;
 import io.lionweb.lioncore.java.serialization.JsonSerialization;
 import io.lionweb.lioncore.java.serialization.SerializationProvider;
--- a/core/src/integrationTest/java/io/lionweb/lioncore/java/testset/ATestset.java
+++ b/core/src/integrationTest/java/io/lionweb/lioncore/java/testset/ATestset.java
@@ -2,7 +2,7 @@ package io.lionweb.lioncore.java.testset;
 
 import static org.junit.Assert.*;
 
-import io.lionweb.lioncore.java.LionWebVersion;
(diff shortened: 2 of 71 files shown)


dslmeinte:
One question: is it possible, with this approach, to introspect on the version at runtime? I see a `LionWebVersionToken` that has that info. btw: It is possible to reflect on type parameters at runtime, despite type erasure! :)

ftomassetti:
> One question: is it possible, with this approach, to introspect on the version at runtime? I see a `LionWebVersionToken` that has that info. btw: It is possible to reflect on type parameters at runtime, despite type erasure! :)

Yes, but that means we need to both specify the type parameter and pass a value to the constructor

ftomassetti:
> btw: It is possible to reflect on type parameters at runtime, despite type erasure! :)

Oh, good. But doesn't that work just with subclasses? Can you show what you had in mind?

dslmeinte:
> > btw: It is possible to reflect on type parameters at runtime, despite type erasure! :)
> 
> Oh, good. But doesn't that work just with subclasses? Can you show what you had in mind?

I'll try and add it to this branch ASAP

ftomassetti:
> > > btw: It is possible to reflect on type parameters at runtime, despite type erasure! :)
> > 
> > 
> > Oh, good. But doesn't that work just with subclasses? Can you show what you had in mind?
> 
> I'll try and add it to this branch ASAP

Hi Meinte, I wanted to check with you if you can estimate when you will have time to look into this (as 2025 is sneaking on us and we are a bit behind with the release of LW Java 2024.1)

dslmeinte:
> > > > btw: It is possible to reflect on type parameters at runtime, despite type erasure! :)
> > > 
> > > 
> > > Oh, good. But doesn't that work just with subclasses? Can you show what you had in mind?
> > 
> > 
> > I'll try and add it to this branch ASAP
> 
> Hi Meinte, I wanted to check with you if you can estimate when you will have time to look into this (as 2025 is sneaking on us and we are a bit behind with the release of LW Java 2024.1)

Sorry: forgot about this (or rather: I didn't put it on some kind of TODO queue) :(

Here's a link: https://stackoverflow.com/questions/1942644/get-generic-type-of-java-util-list
I intended to add an actually representative unit test to `lionweb-java`, but I still have to get around to that.

Issue #172, opened 2024-12-06
Make LionCore M3 and builtins language definitions have version-dependent IDs 


ftomassetti:
See https://github.com/LionWeb-io/specification/issues/279

- [x] Implement version specific IDs
- [x] use resolveInfo and not ID for pointers to built-ins

PR #176, opened 2024-12-13
Validation verify there are no circular SDTs

ftomassetti:
Implements #173

PR #177, opened 2024-12-13
Adopting LW Version dependant IDs for LionCore and LionCoreBuiltins

ftomassetti:
Initial proposal for making the IDs dependant on the LionWeb Version.

Note that tests will fail as we need to update the version of LionCore stored in the specs.

I kept the IDs for 2023.1 as they were, adding the suffix `-2024_1` to the IDs for version 2024.1.

I also started working on a way to "auto-rewire" references to elements from LionCoreBuiltins using the resolveInfo but without enforcing the referred field to be empty.

Implements #172

### DIFF
--- a/core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
@@ -14,7 +14,14 @@ public class LionCoreBuiltins extends Language {
   /** This is private to prevent instantiation and enforce the Singleton pattern. */
   private LionCoreBuiltins(@Nonnull LionWebVersion lionWebVersion) {
     super(lionWebVersion, "LionCore_builtins");
-    setID("LionCore-builtins");
+    final String versionIDSuffix;
+    if (lionWebVersion != LionWebVersion.v2023_1) {
+      versionIDSuffix = "-" + lionWebVersion.getVersionString().replaceAll("\\.", "_");
+    } else {
+      versionIDSuffix = "";
+    }
+
+    setID("LionCore-builtins" + versionIDSuffix);
     setKey("LionCore-builtins");
     setVersion(lionWebVersion.getVersionString());
     PrimitiveType string = new PrimitiveType(lionWebVersion, this, "String");
@@ -24,21 +31,23 @@ public class LionCoreBuiltins extends Language {
       new PrimitiveType(lionWebVersion, this, "JSON");
     }
 
-    Concept node = new Concept(lionWebVersion, this, "Node").setID("LionCore-builtins-Node");
+    Concept node =
+        new Concept(lionWebVersion, this, "Node").setID("LionCore-builtins-Node" + versionIDSuffix);
     node.setAbstract(true);
 
     Interface iNamed =
-        new Interface(lionWebVersion, this, "INamed").setID("LionCore-builtins-INamed");
+        new Interface(lionWebVersion, this, "INamed")
+            .setID("LionCore-builtins-INamed" + versionIDSuffix);
     iNamed.addFeature(
         Property.createRequired(lionWebVersion, "name", string)
-            .setID("LionCore-builtins-INamed-name")
+            .setID("LionCore-builtins-INamed-name" + versionIDSuffix)
             .setKey("LionCore-builtins-INamed-name"));
 
     this.getElements()
         .forEach(
             e -> {
               if (e.getID() == null) {
-                e.setID("LionCore-builtins-" + e.getName());
+                e.setID("LionCore-builtins-" + e.getName() + versionIDSuffix);
               }
               if (e.getKey() == null) {
                 e.setKey("LionCore-builtins-" + e.getName());
--- a/core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
@@ -15,11 +15,10 @@ public class LionCoreBuiltins extends Language {
   private LionCoreBuiltins(@Nonnull LionWebVersion lionWebVersion) {
     super(lionWebVersion, "LionCore_builtins");
     final String versionIDSuffix;
-    if (lionWebVersion != LionWebVersion.v2023_1) {
-      versionIDSuffix = "-" + lionWebVersion.getVersionString().replaceAll("\\.", "_");
-    } else {
-      versionIDSuffix = "";
-    }
+    versionIDSuffix =
+        lionWebVersion != LionWebVersion.v2023_1
+            ? "-" + lionWebVersion.getVersionString().replaceAll("\\.", "_")
+            : "";
 
     setID("LionCore-builtins" + versionIDSuffix);
     setKey("LionCore-builtins");
--- a/core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
@@ -17,7 +17,7 @@ public class LionCoreBuiltins extends Language {
     final String versionIDSuffix;
     versionIDSuffix =
         lionWebVersion != LionWebVersion.v2023_1
-            ? "-" + lionWebVersion.getVersionString().replaceAll("\\.", "_")
+            ? "-" + lionWebVersion.getVersionString().replaceAll("\\.", "-")
             : "";
 
     setID("LionCore-builtins" + versionIDSuffix);
--- a/core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
@@ -1,6 +1,7 @@
 package io.lionweb.lioncore.java.language;
 
 import io.lionweb.lioncore.java.LionWebVersion;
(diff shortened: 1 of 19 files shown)


core/src/main/java/io/lionweb/lioncore/java/serialization/NodePopulator.java
```diff
@@ -98,19 +128,28 @@ private void populateNodeReferences(
...
-                            case THROW_ERROR:
-                              throw new DeserializationException(
-                                  "Unable to resolve reference to "
-                                      + entry.getReference()
-                                      + " for feature "
-                                      + serializedReferenceValue.getMetaPointer());
+
+                          // For LionCore Builtins, we want to automatically update references,
+                          // using Resolve Info
+                          Node autoresolvedElement = autoResolveMap.get(entry.getResolveInfo());
```
enikao, line 134: Why only if `entry.reference != null`? We said for a multi-LW-version compatible language, we'd serialize with only the resolveInfo, no reference.

ftomassetti: You are right, we should do this auto-resolve also when entry.reference is null. I changed that.

Regarding your second point, my understanding was that we had the option of looking at the reference field, and, if and only if that is not set or if it points to a node that cannot be found, we auto-rewire it to the element of the builtins that is available in the current version. So I do not expect `referred` to be necessarily null and in general I would populate it, as it would make the reference behaves "normally" in most situations, with the special behavior being relevant only during migrations.

ftomassetti:
Rebased after merging #176

core/src/main/java/io/lionweb/lioncore/java/serialization/NodePopulator.java
enikao: I'm unsure about the "right" behavior here. Probably needs some experimentation.

ftomassetti: I added three tests for NodePopulator to show that this would work when:
- We refer to a builtin element with its expected id for this version
- We refer to a builtin element with the id it had in another version (i.e., actually with any "wrong" id)
- We refer to a builtin element without specifying the id but just the name

core/src/test/java/io/lionweb/lioncore/java/serialization/NodePopulatorTest.java
```diff
@@ -0,0 +1,181 @@
...
+                    + "        {\n"
+                    + "          \"reference\": {\n"
+                    + "            \"language\": \"LionCore-M3\",\n"
+                    + "            \"version\": \"2024.1\",\n"
+                    + "            \"key\": \"Property-type\"\n"
+                    + "          },\n"
+                    + "          \"targets\": [\n"
+                    + "            {\n"
+                    + "              \"resolveInfo\": \"Boolean\",\n"
+                    + "              \"reference\": \"LionCore-builtins-Boolean\"\n"
```
enikao, line 53: This is the id of LW-2023.1 Boolean, isn't it? Then, according to https://github.com/LionWeb-io/specification/issues/279 _Decision: Variant cc_, this should fail.

In any case, another test case where the node id and the resolveInfo contradict each other might be useful.

dslmeinte (approved):

core/src/test/java/io/lionweb/lioncore/java/serialization/NodePopulatorTest.java
ftomassetti: Right, I corrected this test to use the ID for LW2024.1 Boolean. I then modified another test to use the ID for LW2023.1 Boolean and I verified that cause an exception

core/src/main/java/io/lionweb/lioncore/java/language/LionCoreBuiltins.java
```diff
@@ -14,7 +14,14 @@ public class LionCoreBuiltins extends Language {
   /** This is private to prevent instantiation and enforce the Singleton pattern. */
   private LionCoreBuiltins(@Nonnull LionWebVersion lionWebVersion) {
     super(lionWebVersion, "LionCore_builtins");
-    setID("LionCore-builtins");
+    final String versionIDSuffix;
```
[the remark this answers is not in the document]
ftomassetti: Sure, I modified it accordingly

core/src/test/java/io/lionweb/lioncore/java/serialization/NodePopulatorTest.java
```diff
@@ -50,7 +51,7 @@ public void populateReferenceToBuiltinsValueWithCorrectID() {
                     + "          \"targets\": [\n"
                     + "            {\n"
                     + "              \"resolveInfo\": \"Boolean\",\n"
```
enikao, line 53: According to https://lionweb.io/specification/2024.1/metametamodel/metametamodel.html#predefined-builtins-keys, the _resolveInfo_  should be `LionWeb.LionCore_builtins.Boolean`

ftomassetti: Thank you, I fixed it

core/src/main/java/io/lionweb/lioncore/java/serialization/NodePopulator.java
```diff
@@ -56,7 +56,9 @@ class NodePopulator {
         .getElements()
         .forEach(
             element -> {
-              autoResolveMap.put(element.getName(), element);
+              // See
+              // https://lionweb.io/specification/2024.1/metametamodel/metametamodel.html#predefined-builtins-keys
+              autoResolveMap.put("LionWeb.LionCore_builtins." + element.getName(), element);
```
enikao, line 61: Maybe do the same for M3 elements?
https://lionweb.io/specification/2024.1/metametamodel/metametamodel.html#predefined-m3-keys

ftomassetti:
@enikao 
Following up on the discussion during the check-in in the meeting on the 10th of Jan, I think we need to update IDs like this one:

https://github.com/LionWeb-io/specification/blob/980dd6a8ba5fc4b97d3b53233c09e2bda090c347/2024.1/metametamodel/lioncore.json#L15

As it should be different from:

https://github.com/LionWeb-io/specification/blob/980dd6a8ba5fc4b97d3b53233c09e2bda090c347/2023.1/metametamodel/lioncore.json#L11

enikao:
> @enikao Following up on the discussion during the check-in in the meeting on the 10th of Jan, I think we need to update IDs like this one:

_Technically_ These two are independent, so if you import them in the same system you system should take care of remapping already used ids (https://github.com/LionWeb-io/specification/issues/70). But I agree it's way easier to adjust them in the examples. Could you create a PR?

ftomassetti:
> > @enikao Following up on the discussion during the check-in in the meeting on the 10th of Jan, I think we need to update IDs like this one:
> 
> _Technically_ These two are independent, so if you import them in the same system you system should take care of remapping already used ids ([LionWeb-io/specification#70](https://github.com/LionWeb-io/specification/issues/70)). But I agree it's way easier to adjust them in the examples. Could you create a PR?

PR created: https://github.com/LionWeb-io/specification/pull/325

core/src/main/java/io/lionweb/lioncore/java/model/ClassifierInstanceUtils.java
```diff
@@ -252,4 +250,40 @@ public static void addChild(
...
     _this.addChild(containment, child);
   }
+
+  public static ReferenceValue referenceTo(@Nonnull LanguageEntity<?> _this) {
+    // Unfortunately we cannot refer to LionCore and LionCoreBuiltins as this method get called
+    // during their initialization
+    if (_this.getLanguage() != null
+        && "LionCore_M3".equals(_this.getLanguage().getName())
+        && _this.getLionWebVersion() == LionWebVersion.v2024_1) {
+      return new ReferenceValue(_this, "LionWeb.LionCore_M3." + _this.getName());
```
enikao, line 260: It feels like these magic strings appear several times over the PR. Maybe consolidate them?

core/src/main/java/io/lionweb/lioncore/java/self/LionCore.java
```diff
@@ -164,9 +164,13 @@ private LionCore() {
 
   public static @Nonnull Language getInstance(@Nonnull LionWebVersion lionWebVersion) {
     Objects.requireNonNull(lionWebVersion, "lionWebVersion should not be null");
+    String versionIDSuffix = "";
+    if (lionWebVersion != LionWebVersion.v2023_1) {
+      versionIDSuffix = "-" + lionWebVersion.getVersionString().replaceAll("\\.", "-");
```
enikao, line 169: Also feels like some magical string adjustment that's implemented in several places. Consolidate?

core/src/main/java/io/lionweb/lioncore/java/model/ClassifierInstanceUtils.java
ftomassetti: I will save these constants in a new class, called Autoresolve

core/src/main/java/io/lionweb/lioncore/java/self/LionCore.java
ftomassetti: I created a method in IdUtils to contain this logic

ftomassetti:
Given this has been approved by @dslmeinte and I handled all the issues reported by @enikao I will proceed to merge and release a first version of LW Java with support for 2024.1.

Thank you both for the reviews!

PR #178, opened 2025-01-11
Speed up DynamicNode.equals

ftomassetti:
Currently DynamicNode.equals compare the parent and the concept and such comparisons can be expensive. We can just check their IDs instead, when they are present

### DIFF
--- a/core/src/main/java/io/lionweb/lioncore/java/model/impl/DynamicNode.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/model/impl/DynamicNode.java
@@ -96,7 +96,7 @@ public class DynamicNode extends DynamicClassifierInstance<Concept>
   }
 
   private static boolean shallowNodeEquality(@Nullable Node node1, @Nullable Node node2) {
-    if (node1 == null || node2 == null) {
+    if (node1 == null && node2 == null) {
       return true;
     }
     if (node1 != null && node2 != null && node1.getID() != null) {
--- a/core/src/main/java/io/lionweb/lioncore/java/model/impl/DynamicNode.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/model/impl/DynamicNode.java
@@ -5,6 +5,7 @@ import io.lionweb.lioncore.java.model.HasSettableParent;
 import io.lionweb.lioncore.java.model.Node;
 import java.util.*;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/lionweb/lioncore/java/model/impl/DynamicNode.java
```diff
@@ -70,14 +71,40 @@ public boolean equals(Object o) {
...
+              List<Node> nodes2 = containments2.get(containmentName);
+              return nodes1.size() == nodes2.size()
+                  && IntStream.range(0, nodes1.size())
+                      .allMatch(i -> shallowNodeEquality(nodes1.get(i), nodes2.get(i)));
+            });
+  }
+
+  private static boolean shallowNodeEquality(@Nullable Node node1, @Nullable Node node2) {
+    if (node1 == null || node2 == null) {
+      return true;
```
dslmeinte, line 100: Should only return `true` when both are `null`, right?

core/src/main/java/io/lionweb/lioncore/java/model/impl/DynamicNode.java
```diff
@@ -70,14 +71,40 @@ public boolean equals(Object o) {
     }
     DynamicNode that = (DynamicNode) o;
     return Objects.equals(id, that.id)
-        && Objects.equals(parent, that.parent)
```
dslmeinte, line 73: Are these changes checked by unit tests, potentially implicitly through `assertEquals`?

dslmeinte (approved):

core/src/main/java/io/lionweb/lioncore/java/model/impl/DynamicNode.java
ftomassetti: Right!

core/src/main/java/io/lionweb/lioncore/java/model/impl/DynamicNode.java
ftomassetti: I added a test covering the issue you pointed out above. In DynamicNodeTest we had other tests for equality, so we should have a decent coverage for this method

core/src/main/java/io/lionweb/lioncore/java/model/impl/DynamicNode.java
dslmeinte: But if exactly one of `node1`, `node2` is `null`, then this method falls through to `Objects.equals(node1, node2)` again. Functionally that's the same (I guess), but we could add handling this case easily here, and would improve understandability IMHO.

PR #179, opened 2025-02-05
Version configuration

ftomassetti:
Minor changes to ensure version coherence

### DIFF
--- a/core/src/main/java/io/lionweb/lioncore/java/language/LanguageEntity.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/language/LanguageEntity.java
@@ -33,6 +33,10 @@ public abstract class LanguageEntity<T extends M3Node> extends M3Node<T>
       @Nullable String name,
       @Nonnull String id) {
     this(lionWebVersion, language, name);
+    if (language != null && lionWebVersion != language.getLionWebVersion()) {
+      throw new IllegalArgumentException(
+          "The specified lionWebVersion is not the same as the LionWebVersion of the language");
+    }
     this.setID(id);
   }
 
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/lionweb/lioncore/java/language/LanguageEntity.java
```diff
@@ -33,6 +33,10 @@ public LanguageEntity(
       @Nullable String name,
       @Nonnull String id) {
     this(lionWebVersion, language, name);
+    if (language != null && lionWebVersion != language.getLionWebVersion()) {
```
dslmeinte, line 36: Isn't administrating the LionWeb version **only** at the level of `Language` easier, and less prone to errors (and extra syntactic noise)?

dslmeinte (approved):

core/src/main/java/io/lionweb/lioncore/java/language/LanguageEntity.java
ftomassetti: Yes, but if we created an element without inserting immediately in a Language (or if the language has not the LionWebVersion set) then we do not know which LionWebVersion we should refer to and this is why we have a field in this node.

It is not ideal, but I think that for removing the need to have the lionWebVersion as a field in all the M3 elements we should enforce the nodes to be know their parent at construction time.

dslmeinte: The only way you can have a consistent version in an entire language, is for all its elements to have a version field, and then check as soon as possible whether that's consistent. That might not be trivial, e.g. in the case of having a property with a reference to datatype as the built-in primitive types are dependent on the LionWeb-version. In the end, we might need to validate the entire language anyway after all.

But for now, this is OK.

ftomassetti: That is a good point: we have a class called LanguageValidator. We should add check there to verify the LionWeb Version is the same everywhere. I will add a task for this (https://github.com/LionWeb-io/lionweb-java/issues/180)

PR #183, opened 2025-02-13
Derive abstract property of a LW concept from Ecore isAbstract

Ulyana-F1re:
I find it strange that `abstract` property of a concept is set to `false` in the current version. A metamodel can have an abstract class/concept, so this property should be derived from the Ecore source.

### DIFF
--- a/emf/src/test/java/io/lionweb/lioncore/java/emf/EMFMetamodelImporterTest.java
+++ b/emf/src/test/java/io/lionweb/lioncore/java/emf/EMFMetamodelImporterTest.java
@@ -25,7 +25,7 @@ public class EMFMetamodelImporterTest {
     Language language = languages.get(0);
     assertEquals("library", language.getName());
 
-    assertEquals(5, language.getElements().size());
+    assertEquals(6, language.getElements().size());
 
     Concept book = (Concept) language.getElementByName("Book");
     assertNull(book.getExtendedConcept());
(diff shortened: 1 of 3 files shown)


dslmeinte (approved):
I know it's trivial, but can you add/expand a unit test for this?

emf/src/test/java/io/lionweb/lioncore/java/emf/EMFMetamodelImporterTest.java
```diff
@@ -57,19 +57,22 @@ public void importLibraryExample() throws IOException {
     assertEquals(true, bookAuthor.isRequired());
     assertEquals(false, bookAuthor.isMultiple());
 
+    Concept namedElement = (Concept) language.getElementByName("NamedElement");
+    assertTrue(namedElement.isAbstract());
```
dslmeinte, line 61: This statement sort-of implies it might be good to `assertFalse(<C>.isAbstract())` for the other (concrete) concepts. But that's partly my OCD speaking ;)

dslmeinte (approved):

emf/src/test/java/io/lionweb/lioncore/java/emf/EMFMetamodelImporterTest.java
Ulyana-F1re: yes, those were already there

dslmeinte: Ah, apologies: apparently I suffer from OCD ánd blindness :D

PR #186, opened 2025-02-14
Merge DataTypesMapping into ConceptsToEClassMapping

Ulyana-F1re:
Mappings between Ecore data types and Lionweb data types should be processes in the same way as mappings between Ecore classes and Lionweb concepts. For example, searching for a data type in a EPackage that the current one depends on - was not taken into account in the old DataTypesMapping.


emf/src/main/java/io/lionweb/lioncore/java/emf/mapping/ConceptsToEClassesMapping.java
```diff
@@ -137,6 +145,84 @@ public EClassifier getCorrespondingEClass(Classifier type) {
...
+                "Cannot find corresponding Enumeration for EEnum "
+                        + eEnum.getName()
+                        + " in EPackage "
+                        + eEnum.getEPackage().getName());
+      }
+      processEPackage(eEnum.getEPackage());
+    }
+    if (!eEnumsToEnumerations.containsKey(eEnum)) {
+      throw new IllegalStateException(
+              "Cannot find corresponding Interface for EClass " + eEnum.getName());
```
dslmeinte, line 161: This error message seems incorrect, since we're dealing with enumerations.

dslmeinte, line 161: This error message seems incorrect, since we're dealing with enumerations. (The implementation of `getCorrespondingEEnum` has a correct one.)

dslmeinte, line 161: This error message seems incorrect, since we're dealing with enumerations. (The implementation of `getCorrespondingEEnum` has a correct one.)

dslmeinte (approved):
(but see one remark)

dslmeinte:
> It would be nice to add testing for dependent EPackages, but it will require creating a Registry of the used Ecore packages, for which we will need also a genmodel and the corresponding generated Java code (the factory and the implementation). I am not sure whether we want to have all these in the test resources. Or do we?

The amount of “stuff” is much less of an issue than the effort required to set that up is, or the time saved by having that.

enikao:
> It would be nice to add testing for dependent EPackages, but it will require creating a Registry of the used Ecore packages, for which we will need also a genmodel and the corresponding generated Java code (the factory and the implementation). I am not sure whether we want to have all these in the test resources. Or do we?

Are you sure we need all of that if we set up the Registry dynamically? Ecore _can_ load run-time languages (there's even an option in the UI somewhere).

emf/src/main/java/io/lionweb/lioncore/java/emf/EMFMetamodelExporter.java
```diff
@@ -91,7 +86,15 @@ private void createEClasses(Language language, EPackage ePackage) {
                 eEnum.setName(enumeration.getName());
 
                 ePackage.getEClassifiers().add(eEnum);
-                dataTypeMapping.registerMapping(eEnum, enumeration);
+                conceptsToEClassesMapping.registerMapping(enumeration, eEnum);
```
dslmeinte, line 89: So, this is now more of a `languageEntitiesToEElementsMapping`, right?

emf/src/main/java/io/lionweb/lioncore/java/emf/mapping/ConceptsToEClassesMapping.java
```diff
@@ -10,7 +10,10 @@
 import java.util.Objects;
 import javax.annotation.Nonnull;
 import org.eclipse.emf.ecore.*;
+import org.eclipse.emf.ecore.xml.type.XMLTypePackage;
+import org.eclipse.emf.ecore.xml.type.impl.XMLTypePackageImpl;
 import org.jetbrains.annotations.Nullable;
+import sun.jvm.hotspot.HelloWorld;
 
 public class ConceptsToEClassesMapping {
```
dslmeinte, line 18: This is now a `LanguageEntityToEElementsMapping`, right?

dslmeinte (approved):
but see the comment

dslmeinte (approved):

emf/src/main/java/io/lionweb/lioncore/java/emf/EMFMetamodelImporter.java
```diff
@@ -54,26 +50,26 @@ public Language importEPackage(EPackage ePackage) {
           Interface iface = new Interface(metamodel, eClass.getName());
           setIDAndKey(iface, ePackage.getName() + "-" + iface.getName());
           metamodel.addElement(iface);
-          conceptsToEClassesMapping.registerMapping(iface, eClass);
+          entitiesToEElementsMapping.registerMapping(iface, eClass);
         } else {
           Concept concept = new Concept(metamodel, eClass.getName());
           setIDAndKey(concept, ePackage.getName() + "-" + concept.getName());
-          concept.setAbstract(false);
+          concept.setAbstract(((EClass) eClassifier).isAbstract());
```
ftomassetti, line 57: I think here you could use `eClass`, defined at line 48

ftomassetti (approved):
I added a very minor comment. Sorry for keeping this PR waiting: I did not notice you already addressed the comments by @dslmeinte

emf/src/main/java/io/lionweb/lioncore/java/emf/EMFMetamodelImporter.java
Ulyana-F1re: Yes, indeed. I updated it in the next PR: https://github.com/LionWeb-io/lionweb-java/pull/190

Issue #187, opened 2025-02-14
Improve the DynamicNode equals


tiagobstr:
# Description

The current equals implementation compares the entire object including the containments.

```java
@Override
  public boolean equals(Object o) {
    if (this == o) {
      return true;
    }
    if (!(o instanceof DynamicNode)) {
      return false;
    }
    DynamicNode that = (DynamicNode) o;
    return Objects.equals(id, that.id)
        && Objects.equals(parent, that.parent)
        && Objects.equals(concept, that.concept)
        && Objects.equals(propertyValues, that.propertyValues)
        && Objects.equals(containmentValues, that.containmentValues)
        && Objects.equals(referenceValues, that.referenceValues)
        && Objects.equals(annotations, that.annotations);
  }
```

dslmeinte:
> I'm not 100 % sure that's what we want -- by LionWeb definition, the `id` is the identity of a node. If two nodes have the same id, they are identical, thus also must be equal. If we have two nodes with the same id in the same context, that's an error. If they are in different contexts (e.g. we compare two variants of a common "base"), we need special comparison anyways.

A name like `alsoStructurallyEquals` or `deepEquals` seems more accurate for this method(’s implementation).

ftomassetti:
Solved through #188

PR #188, opened 2025-02-17
Improving the Dynamic node equals

tiagobstr:
## Description

- #187 
- The method now uses the helper methods shallowClassifierInstanceEquality, shallowReferenceEquality, and shallowAnnotationsEquality to compare the respective fields;
- Proposing the change since while working on a project the equals method was caught in a loop then throwing a stack overflow error.

## Testing 

- Ensured no tests are broken;
- Added equality tests for references and annotations.
 
## Checklist

- [X] The correct base branch is being used;
- [x]  All active GitHub checks for tests and formatting are passing;

### DIFF
--- a/core/src/main/java/io/lionweb/lioncore/java/model/impl/DynamicNode.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/model/impl/DynamicNode.java
@@ -1,9 +1,7 @@
 package io.lionweb.lioncore.java.model.impl;
 
 import io.lionweb.lioncore.java.language.*;
-import io.lionweb.lioncore.java.model.ClassifierInstance;
-import io.lionweb.lioncore.java.model.HasSettableParent;
-import io.lionweb.lioncore.java.model.Node;
+import io.lionweb.lioncore.java.model.*;
 import java.util.*;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/lionweb/lioncore/java/model/impl/DynamicNode.java
```diff
@@ -76,8 +78,31 @@ && shallowClassifierInstanceEquality(parent, that.parent)
...
+        .allMatch(
+            referenceName -> {
+              List<ReferenceValue> references1 = reference1.get(referenceName);
+              List<ReferenceValue> references2 = reference2.get(referenceName);
+              return references1.size() == references2.size()
+                  && IntStream.range(0, references1.size())
+                      .allMatch(
+                          i ->
+                              Objects.equals(
+                                      references1.get(i).getResolveInfo(),
```
dslmeinte, line 100: Does `resolveInfo` *need* to be equal? I think only the `ID`s need to match.

dslmeinte (changes requested):
Do we need more unit tests to cover this?

core/src/main/java/io/lionweb/lioncore/java/model/impl/DynamicNode.java
ftomassetti: This is a good point. I am not sure if here we should check strict equality (but then we may want to do a deeper comparison for all aspects of nodes) or to consider "equivalence" instead. 

However I think resolveInfo may be important for this scenario: when we compare two references _not yet solved_ then these two references would both have the referred ID to null, while the resolveInfo may point to different things. In textual languages we may first parse code producing an AST (setting resolveInfo) and then resolve the references (consuming the resolving and set the referred ID). So there is a state when all references are not solved. If we looked only to the referred ID they would all be the same, while eventually they will be solved and point out to different elements

dslmeinte: We're talking instances of `ReferenceValue` here, right? So, we should anyway be able to compare the referred IDs and that's enough. The `resolveInfo`s are not guaranteed to be either not `null`, nor equal for equal reference targets. So, it should really only be a fallback: if refererred IDs are both `null`, then you can assume structural equality if the `resolveInfo`s are equal.

ftomassetti: Right, makes sense

tiagobstr:
@ftomassetti  @dslmeinte I went with giving more importance to what the references point too (referredID) and only using the resolvedInfo if and only if they are not solved. Let me know what you think.

dslmeinte:
> It looks good to me, thank you @tiagobstr . If you could perhaps add 1 or 2 tests I think this PR would be just perfect

+1, especially w.r.t. unit tests

ftomassetti:
Thank you Tiago!

PR #190, opened 2025-02-25
Ensure that proper Lionweb version is set up through the whole language

Ulyana-F1re:
This update ensures that, when importing/exporting from Ecore meta-model and serializing a LionWeb version, we can choose a LionWeb version and have it set up consistently through all language elements.

### DIFF
--- a/core/src/test/java/io/lionweb/lioncore/java/serialization/SerializationOfLibraryTest.java
+++ b/core/src/test/java/io/lionweb/lioncore/java/serialization/SerializationOfLibraryTest.java
@@ -8,13 +8,18 @@ import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
 import io.lionweb.lioncore.java.LionWebVersion;
 import io.lionweb.lioncore.java.language.Concept;
+import io.lionweb.lioncore.java.language.Language;
 import io.lionweb.lioncore.java.language.Property;
 import io.lionweb.lioncore.java.model.ClassifierInstanceUtils;
 import io.lionweb.lioncore.java.model.Node;
+import java.io.File;
+import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.util.List;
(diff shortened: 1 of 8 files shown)


core/src/test/java/io/lionweb/lioncore/java/serialization/SerializationOfLibraryTest.java
```diff
@@ -105,4 +110,39 @@ public void deserializeLanguageWithDuplicateIDs() {
...
+
+  @Rule public TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+  @Test
+  public void reserializeJsonFile() {
+    InputStream inputStream =
+        this.getClass().getResourceAsStream("/serialization/library-language.json");
+    JsonElement jsonElement = JsonParser.parseReader(new InputStreamReader(inputStream));
+    JsonSerialization jsonSerialization =
+        SerializationProvider.getStandardJsonSerialization(LionWebVersion.v2023_1);
```
dslmeinte, line 122: So, only version 2023.1 is tested, effectively? Can you duplicate/parametrize these tests to also check 2024.1?

dslmeinte (approved):
Approved, but see 1 comment

core/src/test/java/io/lionweb/lioncore/java/serialization/SerializationOfLibraryTest.java
Ulyana-F1re: Yes, all tests here use 2023.1 as this is the version used in `library-language.json`. If we want to parameterize these tests, we would need to create a new version of `library-language.json` for every new LionWeb version (it is mostly useful to test for an old version and a current version).

ftomassetti: Makes sense. I will go ahead and merge it as it has been marked as approved

PR #195, opened 2025-03-04
Minor refining of AbstractSerialization

### LINKED ISSUE #189: Fix useless check on Objects.requireNotNull
In AbstractSerialization we have these checks:
```
  private String serializePropertyValue(@Nonnull DataType dataType, @Nullable Object value) {
    Objects.requireNonNull(dataType == null, "cannot serialize property when the dataType is null");
    Objects.requireNonNull(
        dataType.getID() == null, "cannot serialize property when the dataType.ID is null");
   ...
  }
```

The calls to `Objects.requireNonNull` should be corrected.

ftomassetti:
Fix #189 and apply minor refining, like:
- using method references instead of lambdas
- removing unreachable code


dslmeinte (approved):

ftomassetti:
Thank you @dslmeinte !

PR #197, opened 2025-04-04
Support FlatBuffers serialization of partitions

ftomassetti:
Until now, FlatBuffers serialization has not been working for partitions (i.e., for nodes with a null parent). We fix that.

We also sneak in some changes into a method for removing invalid characters from a string and turning it into a valid ID.

### DIFF
--- a/core/src/main/java/io/lionweb/lioncore/java/utils/IdUtils.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/utils/IdUtils.java
@@ -6,6 +6,6 @@ public class IdUtils {
   }
 
   public static String cleanString(String string) {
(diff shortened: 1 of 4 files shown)


dslmeinte (changes requested):
It'd be good to have some unit tests that validate these changes. Right now, I can't really ascertain their validity — at least not without diving into everything downstream of `cleanString` and `serializeNodes`.

ftomassetti:
> It'd be good to have some unit tests that validate these changes. Right now, I can't really ascertain their validity — at least not without diving into everything downstream of cleanString and serializeNodes.

Right, I added the corresponding tests

dslmeinte (approved):

PR #204, opened 2025-04-12
Supporting more APIs from the LionWeb Repository

ftomassetti:
We introduce the AdditionalAPIClient, DBAdminAPIClient and the InspectionAPIClient and corresponding implementation.
We test all of them with functional tests, besides the call for creating a database, as it has no visible effects.

This PR depends on changes in the LionWeb Repository to the bulk import operation: https://github.com/LionWeb-io/lionweb-repository/pull/110 .

Note that in our functional tests we use the commit provided as part of that PR to the LionWeb Repository.


extensions/src/main/java/io/lionweb/serialization/extensions/ExtendedLionWebRepoClient.java
```diff
@@ -0,0 +1,231 @@
...
+    }
+  }
+
+  @Override
+  public List<NodeInfo> getNodeTree(List<String> nodeIDs, @Nullable Integer depthLimit)
+      throws IOException {
+    if (nodeIDs.isEmpty()) {
+      return Collections.emptyList();
+    }
+    String url = protocol.value + "://" + hostname + ":" + port + "/additional/getNodeTree";
```
enikao, line 92: This feels like an invitation for security issues. Isn't there a safer API on e.g. `HttpUrl`?

extensions/src/main/java/io/lionweb/serialization/extensions/NodeInfo.java
```diff
@@ -0,0 +1,57 @@
+package io.lionweb.serialization.extensions;
+
+import java.util.Objects;
+import org.jetbrains.annotations.Nullable;
+
+public final class NodeInfo {
```
enikao, line 6: Maybe add a bit of docs explaining the difference between `NodeInfo`, `SerializedNode`, etc?

repo-client/src/main/java/io/lionweb/repoclient/api/ClassifierKey.java
```diff
@@ -0,0 +1,47 @@
+package io.lionweb.repoclient.api;
+
+import java.util.Objects;
+
+public final class ClassifierKey {
```
enikao, line 5: I'd expect to also have something like a `LanguageKey`, as a language is identified by the tuple _(key, version)_.

repo-client/src/main/java/io/lionweb/repoclient/api/HistorySupport.java
```diff
@@ -0,0 +1,14 @@
+package io.lionweb.repoclient.api;
+
+public enum HistorySupport {
+  Disabled,
+  Enabled;
+
+  public static HistorySupport fromBoolean(boolean value) {
```
enikao, line 7: `@NotNull`?

extensions/src/main/java/io/lionweb/serialization/extensions/AdditionalAPIClient.java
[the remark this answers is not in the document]
ftomassetti: I am using the same names used in the LionWeb Repository (https://github.com/LionWeb-io/lionweb-repository/blob/6fec1602c00cdd8bdfaf710f746e3f918c9a0b8d/packages/additionalapi/src/main.ts#L38). This group contains two different APIs that are unrelated. The only thing in common is being useful stuff, not being required by the specs. They are practical things that proved to be useful in practice. I would not know what name could be better, and so I would default to keep the same name as in the repository

extensions/src/main/java/io/lionweb/serialization/extensions/CompressionSupport.java
```diff
@@ -0,0 +1,64 @@
...
+import okio.Buffer;
+import okio.BufferedSink;
+import okio.GzipSink;
+import okio.Okio;
+
+public class CompressionSupport {
+
+  public static RequestBody considerCompression(RequestBody original, Compression compression)
+      throws IOException {
+    if (compression == Compression.ENABLED) {
```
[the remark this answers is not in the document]
ftomassetti: Will do

repo-client/src/main/java/io/lionweb/repoclient/api/HistorySupport.java
```diff
@@ -0,0 +1,14 @@
+package io.lionweb.repoclient.api;
+
+public enum HistorySupport {
+  Disabled,
```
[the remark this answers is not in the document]
ftomassetti: Right, will change

repo-client/src/main/java/io/lionweb/repoclient/LionWebRepoClient.java
```diff
@@ -22,17 +24,17 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-public class LionWebRepoClient {
+public class LionWebRepoClient implements BulkAPIClient, DBAdminAPIClient, InspectionAPIClient {
```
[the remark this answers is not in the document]
ftomassetti: I completely revised the implementation, extracting the implementation of each interface in a separate class. I created a base class with the helper methods. I kept this class to act as a facade, providing the same features as before, but with the benefit of having the implementation spread in several classes

repo-client/src/main/java/io/lionweb/repoclient/api/ClassifierKey.java
ftomassetti: The shape of this object depends on the values returned by the LionWeb repository, so we cannot change it here without first changing it in the repository

extensions/src/main/java/io/lionweb/serialization/extensions/NodeInfo.java
ftomassetti: I added a comment for this class

ftomassetti:
Thank you both for your review: the code is now significantly better.

Issue #206, opened 2025-04-17
Work on performance of serialization


ftomassetti:
# Example

I defined a large language and serialized it to JSON. It is ~12MB and it contains thousands of concepts.

# Test

I defined the following test class:

```
package io.lionweb.lioncore.java.serialization;

import static org.junit.Assert.assertEquals;
import static org.junit.Assert.assertTrue;

import io.lionweb.lioncore.java.LionWebVersion;
import io.lionweb.lioncore.java.model.Node;
import java.io.*;
import java.io.InputStream;
import java.util.ArrayList;
import java.util.List;
import java.util.stream.Collectors;
import org.junit.Test;

public class PerformanceTestOnSerialization {

  @Test
  public void deserializeLargeLanguage() {
    InputStream is = this.getClass().getResourceAsStream("/serialization/LargeLanguage.json");
    String json = readInputStreamToString(is);
    JsonSerialization js =
        SerializationProvider.getStandardJsonSerialization(LionWebVersion.v2023_1);

    performanceMeasure(() -> js.deserializeToNodes(json), 250, 350);
  }

  @Test
  public void serializeLargeLanguage() {
    InputStream is = this.getClass().getResourceAsStream("/serialization/LargeLanguage.json");
    String json = readInputStreamToString(is);
    JsonSerialization js =
        SerializationProvider.getStandardJsonSerialization(LionWebVersion.v2023_1);
    List<Node> roots =
        js.deserializeToNodes(json).stream()
            .filter(n -> n.getParent() == null)
            .collect(Collectors.toList());
    assertEquals(2, roots.size());

    // Let's create a separate JsonSerialization, just in case some caches could affect the result
    final JsonSerialization js2 =
        SerializationProvider.getStandardJsonSerialization(LionWebVersion.v2023_1);
    performanceMeasure(
        () -> js2.serializeTreesToJsonElement(roots.get(0), roots.get(1)), 150, 200);
  }

  private String readInputStreamToString(InputStream inputStream) {
    try (BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream))) {
      return reader.lines().collect(Collectors.joining("\n"));
    } catch (IOException e) {
      throw new UncheckedIOException(e);
    }
  }

  private void performanceMeasure(Runnable runnable, long thresholdMin, long thresholdMax) {
    List<Long> elapsedList = new ArrayList<>();
    int N_ITERATIONS = 25;
    int N_TOP_REMOVED = 4;
[...]

PR #207, opened 2025-04-17
Improve serialization performance

ftomassetti:
Issue #206 reports the measurements obtained for the serialization and deserialization of a large file.

* deserialization: 522ms to 545ms
* serialization: 6595ms to 6685ms

With this PR we get:

* deserialization: 217ms to 358ms
* serialization: 123ms to 130ms

Considering the best case, the deserialization is down to 42% the original time and the serialization is down to 2% of the original time (i.e., we got over a 50x improvement).

We do that mainly by ntroducing `SerializationStatus` and adding features to `DeserializationStatus`. These two classes caches expensive operations, like calculating all the features of a classifier.

We also replace some calls to `Objects.requireNonNull` with ifs, to avoid expensive message calculations.


core/src/main/java/io/lionweb/lioncore/java/serialization/AbstractSerialization.java
```diff
@@ -220,7 +227,7 @@ public SerializedChunk serializeNodesToSerializationBlock(
   }
 
   private SerializedClassifierInstance serializeNode(
-      @Nonnull ClassifierInstance<?> classifierInstance) {
+      @Nonnull ClassifierInstance<?> classifierInstance, SerializationStatus serializationStatus) {
```
enikao, line 230: If we hand the `serializationStatus` to every method, isn't it part of the serializer's state? Then, we can just make it a field.

core/src/main/java/io/lionweb/lioncore/java/serialization/DeserializationStatus.java
```diff
@@ -22,10 +20,17 @@
...
  */
 class DeserializationStatus {
-  final List<SerializedClassifierInstance> sortedList;
+  private final List<SerializedClassifierInstance> sortedList;
   final List<SerializedClassifierInstance> nodesToSort;
   final List<ProxyNode> proxies = new ArrayList<>();
-  private LocalClassifierInstanceResolver proxiesInstanceResolver;
+  private final LocalClassifierInstanceResolver proxiesInstanceResolver;
+  private final Set<String> sortedIDs = new HashSet<>();
+  private PrimitiveValuesSerialization primitiveValuesSerialization;
```
enikao, line 28: why not final (all fields)?

core/src/main/java/io/lionweb/lioncore/java/serialization/DeserializationStatus.java
```diff
@@ -35,26 +40,80 @@ class DeserializationStatus {
...
     sortedList = new ArrayList<>();
     nodesToSort = new ArrayList<>(originalList);
     this.proxiesInstanceResolver = new LocalClassifierInstanceResolver();
     this.globalInstanceResolver =
         new CompositeClassifierInstanceResolver(outsideInstancesResolver, proxiesInstanceResolver);
   }
 
+  public Property getProperty(Classifier<?> classifier, MetaPointer metaPointer) {
+    if (!featuresCache.containsKey(classifier)) {
+      featuresCache.put(classifier, new HashMap<>());
```
enikao, line 55: maybe use `putIfAbsent()`? (also other similar places)

core/src/main/java/io/lionweb/lioncore/java/serialization/AbstractSerialization.java
```diff
@@ -164,43 +164,50 @@ public SerializedChunk serializeNodesToSerializationBlock(
       Collection<ClassifierInstance<?>> classifierInstances) {
     SerializedChunk serializedChunk = new SerializedChunk();
     serializedChunk.setSerializationFormatVersion(lionWebVersion.getVersionString());
+    SerializationStatus serializationStatus = new SerializationStatus();
     for (ClassifierInstance<?> classifierInstance : classifierInstances) {
```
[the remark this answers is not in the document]
ftomassetti: Thank you for the suggestion. No I did not think about it: I tried now and, for some reason which I cannot understand it seems to over twice as slow when using parallel...

core/src/main/java/io/lionweb/lioncore/java/serialization/AbstractSerialization.java
ftomassetti: I think it is similar to deserializationStatus: we keep it while serving one request and we pass it to a few internal methods but when the request is satisfied the instance of serializationStatus is released, which means also the caches within serializationStatus are released

core/src/main/java/io/lionweb/lioncore/java/serialization/DeserializationStatus.java
ftomassetti: Right, making them final

core/src/main/java/io/lionweb/lioncore/java/serialization/DeserializationStatus.java
ftomassetti: Right! I will use computeIfAbsent

core/src/main/java/io/lionweb/lioncore/java/serialization/SerializationStatus.java
```diff
@@ -0,0 +1,46 @@
...
+import io.lionweb.lioncore.java.language.Containment;
+import io.lionweb.lioncore.java.language.Property;
+import io.lionweb.lioncore.java.language.Reference;
+import java.util.HashSet;
+import java.util.IdentityHashMap;
+import java.util.List;
+import java.util.Set;
+
+public class SerializationStatus {
+  private Set<String> classifiersConsidered = new HashSet<>();
```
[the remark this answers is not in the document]
ftomassetti: Right, changing it

core/src/main/java/io/lionweb/lioncore/java/serialization/DeserializationStatus.java
```diff
@@ -35,26 +40,65 @@ class DeserializationStatus {
...
     nodesToSort = new ArrayList<>(originalList);
     this.proxiesInstanceResolver = new LocalClassifierInstanceResolver();
     this.globalInstanceResolver =
         new CompositeClassifierInstanceResolver(outsideInstancesResolver, proxiesInstanceResolver);
   }
 
+  public Property getProperty(Classifier<?> classifier, MetaPointer metaPointer) {
+    featuresCache.computeIfAbsent(classifier, c -> new HashMap<>());
+    Map<MetaPointer, Feature<?>> featuresMap = featuresCache.get(classifier);
+    featuresMap.computeIfAbsent(metaPointer, classifier::getPropertyByMetaPointer);
```
enikao, line 56: I think we can even compress this further -- `return featuresMap.computeIfAbsent()` should work.

ftomassetti: Absolutely. We can also compress the previous two lines in one (and similarly for the other two methods)

core/src/main/java/io/lionweb/lioncore/java/serialization/DeserializationStatus.java
```diff
@@ -35,26 +40,66 @@ class DeserializationStatus {
...
+    return (Reference)
+        featuresMap.computeIfAbsent(metaPointer, classifier::getReferenceByMetaPointer);
+  }
+
+  public Object deserializePropertyValue(
+      DataType<?> dataType, String serializedValue, boolean isRequired) {
+    propertyValuesCache.computeIfAbsent(dataType, dt -> new HashMap<>());
+    Map<String, Object> map = propertyValuesCache.get(dataType);
+    String key = serializedValue + "@required@" + isRequired;
+    map.computeIfAbsent(
```
enikao, line 80: combine?

core/src/main/java/io/lionweb/lioncore/java/serialization/SerializationStatus.java
```diff
@@ -0,0 +1,46 @@
...
+  public boolean hasConsideredClassifier(String classifierId) {
+    return classifiersConsidered.contains(classifierId);
+  }
+
+  public void markClassifierAsConsidered(String classifierId) {
+    classifiersConsidered.add(classifierId);
+  }
+
+  public Iterable<Property> allProperties(Classifier<?> classifier) {
+    if (!properties.containsKey(classifier.getID())) {
```
enikao, line 27: combine?

core/src/main/java/io/lionweb/lioncore/java/serialization/DeserializationStatus.java
ftomassetti: Done

core/src/main/java/io/lionweb/lioncore/java/serialization/SerializationStatus.java
ftomassetti: Done

enikao:
Did the map lookup tuning help with the overall performance?

PR #208, opened 2025-04-18
Serialization performance follow-up

ftomassetti:
In this PR we do another couple of refinements for the performance of serialization, based on a concrete case I had to deal with.

The case I considered:

* Used annotations
* Used many instances of many different concepts

So I applied similar changes we did to handle nodes to the handling of annotations. Also, I realized on expensive operation was considering new languages for serialization, so I put some guardrails in place to have adding multiple times the same language. 

In my case this helped improving the performance significantly.


enikao (approved):
I don't really understand what's happening here, but also don't have any remarks on the code.

ftomassetti:
Thank you!

PR #213, opened 2025-05-01
Implement History APIs Client

ftomassetti:
We update the commit used, as we need to the fix https://github.com/LionWeb-io/lionweb-repository/pull/113

~~We also add the validation annotations, as they can be used in cases where the jetbrains' ones cannot.~~

~~E.g.: `@Nonnull List<@Nonnull String> foo()`~~

We add the necessary functional tests.

We also ensure that bulk/store returns the repo version when present.

With this we are supporting all the APIs exposed by the LionWeb Repository. Next step is supporting the delta protocol.


repo-client/src/main/java/io/lionweb/repoclient/LionWebRepoClient.java
```diff
@@ -112,7 +110,7 @@ public LionWebRepoClient(
       @NotNull LionWebVersion lionWebVersion,
       @NotNull String hostname,
       int port,
-      @Nullable String authorizationToken,
+      @Null String authorizationToken,
```
enikao, line 113: I think `@Null` has different meaning. Quoting from [docs](https://docs.oracle.com/javaee/7/api/javax/validation/constraints/Null.html):

> The annotated element must be null.

repo-client/src/main/java/io/lionweb/repoclient/impl/ClientForBulkAPIs.java
```diff
@@ -31,18 +33,18 @@ public ClientForBulkAPIs(RepoClientConfiguration repoClientConfiguration) {
   }
 
   @Override
-  public void createPartitions(List<Node> partitions) throws IOException {
-    createPartitions(
+  public @Nullable Long createPartitions(List<Node> partitions) throws IOException {
```
enikao, line 36: Maybe add some JavaDoc what the return value means?

If I get it right, it means the version of the history. In that case, I'd create an interface that clearly represents that meaning, and have potentially different implementations -- one repo might just count them, another might use git hashes, etc.

ftomassetti: I agree that having a type that more clearly represents the repository version makes sense, so I introduced `RepositoryVersionToken`.

However, I think that an interface with multiple implementations would be overkill at the current stage. I think we can just treat the version as a string. It should work for all kind of version tokens used. I think we can ignore the fact that htis particular implementation of the repository returns a long and just treat it as string, as we do not expect to do math on it or even sort the versions specified.

repo-client/src/main/java/io/lionweb/repoclient/LionWebRepoClient.java
```diff
@@ -141,6 +139,7 @@ public LionWebRepoClient(
     this.inspectionAPIs = new ClientForInspectionAPIs(conf);
     this.dbAdminAPIs = new ClientForDBAdminAPIs(conf);
     this.bulkAPIs = new ClientForBulkAPIs(conf);
+    this.historyAPIs = new ClientForHistoryAPIs(conf);
```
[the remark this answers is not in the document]
ftomassetti: You are right, I am reverting back to the jetbrains' ones

repo-client/src/main/java/io/lionweb/repoclient/LionWebRepoClient.java
ftomassetti: You are right, I am reverting back to the jetbrains' annotations

repo-client/src/main/java/io/lionweb/repoclient/impl/ClientForBulkAPIs.java
enikao: So we don't want `RepositoryVersionToken` to be `Comparable`? I don't have an issue with that, we just should make a _conscious_ decision about that.

repo-client/src/main/java/io/lionweb/repoclient/api/HistoryAPIClient.java
```diff
@@ -0,0 +1,29 @@
...
+
+import io.lionweb.lioncore.java.model.Node;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import org.jetbrains.annotations.NotNull;
+
+public interface HistoryAPIClient {
+  @NotNull
+  List<Node> historyListPartitions(RepositoryVersionToken repoVersion) throws IOException;
```
enikao, line 11: commonly, method names start with a verb.

repo-client/src/main/java/io/lionweb/repoclient/api/RepositoryVersionToken.java
```diff
@@ -0,0 +1,28 @@
+package io.lionweb.repoclient.api;
+
+import java.util.Objects;
+
+/** Value distinguishing a version of a particular repository. */
+public class RepositoryVersionToken {
+  private String token;
```
enikao, line 7: `final`? Or are we on a sufficiently modern version of Java to use `record`?

repo-client/src/main/java/io/lionweb/repoclient/impl/ClientForBulkAPIs.java
ftomassetti: I would start by _not_ making it `Comparable`, as if someone start using things other than longs for identifying versions I would not know how to compare them. If someone signals the needs to have versions comparable, then we can add it

repo-client/src/main/java/io/lionweb/repoclient/api/RepositoryVersionToken.java
ftomassetti: Making the field final. And not, we do not have support for record. We are supporting Java 8, which is very old. At some point we may want to move to Java 11, but I think records came later than that

repo-client/src/main/java/io/lionweb/repoclient/api/HistoryAPIClient.java
ftomassetti: I removed the `history` prefix from this and the `historyRetrieve` (now `retrieve`) method.

Now they have the same name as the variants without history, but given Java supports method overloading and the type of the first parameter (`RepositoryVersionToken`) make it obvious which variants are connected with history, then I do not see a problem with that

repo-client/src/main/java/io/lionweb/repoclient/api/BulkAPIClient.java
```diff
@@ -3,17 +3,21 @@
 import io.lionweb.lioncore.java.model.Node;
 import java.io.IOException;
 import java.util.List;
+import org.jetbrains.annotations.Nullable;
 
 public interface BulkAPIClient {
-  void createPartitions(List<Node> partitions) throws IOException;
+  @Nullable
+  RepositoryVersionToken createPartitions(List<Node> partitions) throws IOException;
```
enikao, line 10: If we separate regular client from history client, do we want to introduce history stuff (`RepositoryVersionToken`) to the regular client?

ftomassetti: We have separate interfaces: one for each module of the repository.

We do that for two reasons:

* because of a previous comment by Meinte on having a more modular implementation
* because in this way, one could introduce _other implementations_ for a subset of the interface

For the implementation currently in this codebase I think it makes sense to support all the interfaces because most methods work both for repository with or without history.

enikao: I was thinking whether we want to have anything history-related in the non-history Interface. We still can implement all interfaces on one class.

ftomassetti: The non-history interface is extended by the history one. We could duplicate the methods such as listPartitions and createPartitions, but then one would need to call different methods depending on the fact they are operating on a repo with history or without. It seems to be over complicated. I thought about that for one week but I could not find a proper solution, so I would lean toward merging this as-is

PR #216, opened 2025-05-09
Configure for publishing to Central Portal

ftomassetti:
This is necessary as we moved from OSSRH to Central Portal, as OSSRH is about to be decommissioned.

Since we did the transition we cannot publish through OSSRH anymore.

I tested this by publishing snapshot (which worked).

I am asking a review to @enikao as it may be relevant for LionWeb MPS.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,7 +2,7 @@ plugins {
     id "com.diffplug.spotless" version "5.1.0"
     id "java"
     id 'net.researchgate.release' version '3.0.2'
-    id("com.vanniktech.maven.publish") version "0.29.0" apply false
+    alias(libs.plugins.vtpublish) apply false
 }
 
 repositories {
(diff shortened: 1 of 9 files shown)


dslmeinte (approved):
LGTM, but then again, I don't know enough about Maven to really judge this. Do you think the migration to Central Portal should also be done for LionWeb MPS?

ftomassetti:
Thank you for the review!
Yes, we need to do the changes to LionWeb MPs as well as the migration
works for the entire namespace and as of now we can only publish through
the new system and not anymore through the old one (one could not keep them
both active at the same time). This why I checked in the chat before
activating the migration for the namespace io.lionweb. The good news is
that the changes needed are minimal: changes to the code as in this PR plus
the generation of a new token for the new system to be used instead of the
old token

Software Architect & Founder at Strumenta (https://strumenta.com)
Technical blog at https://tomassetti.me
GitHub https://github.com/ftomassetti
Twitter @ftomasse
linkedin.com/in/federicotomassetti

enikao:
Thanks for tagging me --> https://github.com/LionWeb-io/lionweb-mps/issues/135

PR #217, opened 2025-05-11
Add support for downloading/uploading entire repositories

ftomassetti:
For backup/restore use cases it is useful to have a command to download an entire repository to a directory or as a zip and then reupload it from there. This feature is introduced into the extensions.

Changes are made to expose lower level operations (besides the existing "normal" operations) which permits to communicate with the repository without being forced to always deserialize nodes (e.g., we can the nodes serialized and we can save them on file, we can then load them from file and, without deserializing them, send them to the repository).

### DIFF
--- a/core/src/main/java/io/lionweb/lioncore/java/serialization/SerializationStatus.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/serialization/SerializationStatus.java
@@ -5,7 +5,6 @@ import io.lionweb.lioncore.java.serialization.data.MetaPointer;
 import io.lionweb.lioncore.java.serialization.data.SerializedChunk;
 import io.lionweb.lioncore.java.serialization.data.SerializedClassifierInstance;
 import io.lionweb.lioncore.java.serialization.data.UsedLanguage;
-
 import java.util.*;
 import java.util.function.Consumer;
 
--- a/core/src/main/java/io/lionweb/lioncore/java/serialization/SerializationStatus.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/serialization/SerializationStatus.java
@@ -69,32 +69,42 @@ public class SerializationStatus {
 
   /** Consider the given classifier and all of its properties to track the used languages. */
   public void consider(Classifier<?> classifier, Consumer<Language> languageConsumer) {
-    if (!hasConsideredClassifier(classifier.getID())) {
-      Objects.requireNonNull(classifier, "A node should have a concept in order to be serialized");
-      Language language = classifier.getLanguage();
-      if (language == null) {
-        throw new NullPointerException(
-            "A Concept should be part of a Language in order to be serialized. Concept "
-                + classifier
-                + " is not");
-      }
-      considerLanguageDuringSerialization(languageConsumer, language);
-      List<Feature<?>> features = classifier.allFeatures();
-      features.forEach(
-          f -> considerLanguageDuringSerialization(languageConsumer, f.getDeclaringLanguage()));
-      features.stream()
-          .filter(f -> f instanceof Property)
-          .map(f -> (Property) f)
-          .forEach(
-              p ->
-                  considerLanguageDuringSerialization(languageConsumer, p.getType().getLanguage()));
-      features.stream()
-          .filter(f -> f instanceof Link)
-          .map(f -> (Link<?>) f)
-          .forEach(
-              l ->
-                  considerLanguageDuringSerialization(languageConsumer, l.getType().getLanguage()));
-      markClassifierAsConsidered(classifier.getID());
+    Objects.requireNonNull(classifier, "A node should have a concept in order to be serialized");
+    if (hasConsideredClassifier(classifier.getID())) {
+      return;
     }
+    Language language = classifier.getLanguage();
+    if (language == null) {
+      throw new IllegalStateException(
+          "A Classifier should be part of a Language in order to be serialized. Classifier "
+              + classifier
+              + " is not");
+    }
+    considerLanguageDuringSerialization(languageConsumer, language);
+    List<Feature<?>> features = classifier.allFeatures();
+    features.forEach(
+        f -> considerLanguageDuringSerialization(languageConsumer, f.getDeclaringLanguage()));
+    features.stream()
+        .filter(f -> f instanceof Property)
+        .map(f -> (Property) f)
+        .forEach(
+            p -> {
+              DataType<?> dt = p.getType();
+              considerLanguageDuringSerialization(languageConsumer, dt.getLanguage());
+              if (dt instanceof StructuredDataType) {
+                StructuredDataType sdt = (StructuredDataType) dt;
+                sdt.getFields()
+                    .forEach(
+                        f ->
+                            considerLanguageDuringSerialization(
+                                languageConsumer, f.getType().getLanguage()));
+              }
+            });
+    features.stream()
+        .filter(f -> f instanceof Link)
+        .map(f -> (Link<?>) f)
+        .forEach(
+            l -> considerLanguageDuringSerialization(languageConsumer, l.getType().getLanguage()));
+    markClassifierAsConsidered(classifier.getID());
   }
 }
--- a/core/src/main/java/io/lionweb/lioncore/java/serialization/SerializationStatus.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/serialization/SerializationStatus.java
@@ -64,6 +64,9 @@ public class SerializationStatus {
     }
   }
 
+  /**
+   * Consider the given classifier and all of its properties to track the used languages.
+   */
   public void consider(Classifier<?> classifier, Consumer<Language> languageConsumer) {
     if (!hasConsideredClassifier(classifier.getID())) {
       Objects.requireNonNull(classifier, "A node should have a concept in order to be serialized");
--- a/core/src/main/java/io/lionweb/lioncore/java/serialization/SerializationStatus.java
+++ b/core/src/main/java/io/lionweb/lioncore/java/serialization/SerializationStatus.java
@@ -1,7 +1,11 @@
 package io.lionweb.lioncore.java.serialization;
 
 import io.lionweb.lioncore.java.language.*;
+import io.lionweb.lioncore.java.serialization.data.MetaPointer;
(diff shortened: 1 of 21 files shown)


ftomassetti:
The failure seems due to itemis cloud returning unauthorized, so I would ignore it for now

core/src/main/java/io/lionweb/lioncore/java/serialization/SerializationStatus.java
```diff
@@ -41,4 +51,50 @@ public void considerLanguageDuringSerialization(Consumer<Language> consumer, Lan
...
+    UsedLanguage ul = new UsedLanguage(metaPointer.getLanguage(), metaPointer.getVersion());
+    if (!usedLanguages.contains(ul)) {
+      usedLanguages.add(ul);
+      serializedChunk.addLanguage(ul);
+    }
+  }
+
+  /** Consider the given classifier and all of its properties to track the used languages. */
+  public void consider(Classifier<?> classifier, Consumer<Language> languageConsumer) {
+    if (!hasConsideredClassifier(classifier.getID())) {
```
enikao, line 72: Maybe invert the condition to have early exit and less nesting.

core/src/main/java/io/lionweb/lioncore/java/serialization/SerializationStatus.java
```diff
@@ -41,4 +51,50 @@ public void considerLanguageDuringSerialization(Consumer<Language> consumer, Lan
...
+      Objects.requireNonNull(classifier, "A node should have a concept in order to be serialized");
```
enikao, line 73: This test can never be reached -- the line above would already NPE

core/src/main/java/io/lionweb/lioncore/java/serialization/SerializationStatus.java
```diff
@@ -41,4 +51,50 @@ public void considerLanguageDuringSerialization(Consumer<Language> consumer, Lan
...
+      Language language = classifier.getLanguage();
+      if (language == null) {
+        throw new NullPointerException(
```
enikao, line 76: Is this the right kind of exception? I'd consider `InvalidStateException` or maybe `IllegalArgumentException`. Or (that would be my preferred direction) introduce specific LionWeb-related exceptions.

core/src/main/java/io/lionweb/lioncore/java/serialization/SerializationStatus.java
```diff
@@ -41,4 +51,50 @@ public void considerLanguageDuringSerialization(Consumer<Language> consumer, Lan
...
+            "A Concept should be part of a Language in order to be serialized. Concept "
+                + classifier
+                + " is not");
+      }
+      considerLanguageDuringSerialization(languageConsumer, language);
+      List<Feature<?>> features = classifier.allFeatures();
+      features.forEach(
+          f -> considerLanguageDuringSerialization(languageConsumer, f.getDeclaringLanguage()));
+      features.stream()
+          .filter(f -> f instanceof Property)
```
enikao, line 86: In theory, if the property is of type _StructuredDataType_, the SDT can have fields with types of other languages that would also need to be considered.

extensions/src/main/java/io/lionweb/repoclient/RepoSerialization.java
```diff
@@ -0,0 +1,261 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+import java.util.zip.ZipOutputStream;
+import kotlin.text.Charsets;
+
+/** This class contains the logic to store and retrieve entire repositories at once. */
+public class RepoSerialization {
+  private int nNodesThreshold = 100_000;
```
enikao, line 30: typo? "nNodes"

extensions/src/main/java/io/lionweb/repoclient/RepoSerialization.java
```diff
@@ -0,0 +1,261 @@
...
+  private TransferFormat transferFormat = TransferFormat.FLATBUFFERS;
+  private Compression compression = Compression.DISABLED;
+
+  /**
+   * Download all the content of the repository accessed by the apiClient into a directory. In this
+   * directory one file per partition is created. The file is in JSON format.
+   */
+  public <C extends RawBulkAPIClient & BulkAPIClient> void downloadRepoAsDirectory(
```
enikao, line 38: This is advanced genericing!

extensions/src/main/java/io/lionweb/repoclient/RepoSerialization.java
```diff
@@ -0,0 +1,261 @@
...
+    }
+  }
+
+  /**
+   * Download all the content of the repository accessed by the apiClient into a zip file. In this
+   * zip file, one entry per partition is created. The entry is in JSON format.
+   */
+  public <C extends RawBulkAPIClient & BulkAPIClient> void downloadRepoAsZip(
+      C apiClient, File zipFile) throws IOException {
+    try (FileOutputStream fos = new FileOutputStream(zipFile);
```
enikao, line 58: I'm not sure about current best practices in Java, but it used to be recommended to have a BufferedStream somewhere in there.

extensions/src/main/java/io/lionweb/repoclient/RepoSerialization.java
```diff
@@ -0,0 +1,261 @@
...
+   */
+  public void simpleUploadDirectoryToRepo(RawBulkAPIClient apiClient, File directory)
+      throws IOException {
+    if (!directory.isDirectory()) {
+      throw new IllegalArgumentException(
+          "Provided file is not a directory: " + directory.getAbsolutePath());
+    }
+
+    for (File file : findJsonFilesRecursively(directory)) {
+      String content = new String(Files.readAllBytes(file.toPath()), Charsets.UTF_8);
```
enikao, line 89: This means your whole file needs to fit in memory twice -- once as string, once as SerializedNodes.

extensions/src/main/java/io/lionweb/repoclient/RepoSerialization.java
```diff
@@ -0,0 +1,261 @@
...
+          continue;
+        }
+
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        int bytesRead;
+        while ((bytesRead = zis.read(buffer)) != -1) {
+          baos.write(buffer, 0, bytesRead);
+        }
+
+        // Write to temp file (needed because deserializeSerializationBlock expects a File)
```
enikao, line 227: This feels like `deserializeSerializationBlock` should be refactored to work on streams

repo-client/src/main/java/io/lionweb/repoclient/LionWebRepoClient.java
```diff
@@ -162,6 +168,27 @@ protected RepoClientConfiguration buildRepositoryConfiguration() {
     return jsonSerialization;
   }
 
+  //
+  // Raw Bulk APIs
+  //
+
+  public @Nullable RepositoryVersionToken rawCreatePartitions(@NotNull String data)
```
enikao, line 175: Should the parameters really be strings? This forces to have them in memory. Maybe we can go with `CharSequence` -- that's implemented by `String`, so it's compatible, but at least leaves the option to use something more efficient.

enikao: Or we directly go to `Stream`.

core/src/main/java/io/lionweb/lioncore/java/serialization/SerializationStatus.java
ftomassetti: Good idea

core/src/main/java/io/lionweb/lioncore/java/serialization/SerializationStatus.java
ftomassetti: Right, moving it on top

core/src/main/java/io/lionweb/lioncore/java/serialization/SerializationStatus.java
ftomassetti: Makes sense, I am making it an InvalidStateException

core/src/main/java/io/lionweb/lioncore/java/serialization/SerializationStatus.java
ftomassetti: You are right, adding it

extensions/src/main/java/io/lionweb/repoclient/RepoSerialization.java
ftomassetti: It stands for number of nodes. I have renamed it to `numberOfNodesThreshold`

extensions/src/main/java/io/lionweb/repoclient/RepoSerialization.java
ftomassetti: Yes :D

extensions/src/main/java/io/lionweb/repoclient/RepoSerialization.java
ftomassetti: Yes, you are right. It should improve performance

extensions/src/main/java/io/lionweb/repoclient/RepoSerialization.java
ftomassetti: Sadly yes. Unfortunately this is the case because we currently cannot create partitions with children: we must create first without children and then add the children in a second call. So we need to process the string into a data structure that we can manipulate to remove children.

extensions/src/main/java/io/lionweb/repoclient/RepoSerialization.java
ftomassetti: This is now https://github.com/LionWeb-io/lionweb-java/issues/218

repo-client/src/main/java/io/lionweb/repoclient/LionWebRepoClient.java
ftomassetti: I tried making it into a CharSequence but eventually we reach methods from okhttp that wants strings and cannot accept CharSequences

Issue #220, opened 2025-06-23
Serialization of languages doesn't mention LionCore-builtins in used languages


dslmeinte:
This was signalled through the integration testing repo: https://github.com/LionWeb-io/lionweb-integration-testing/actions/runs/15831929543/job/44626475550#step:10:37

The specification says: “If the chunk describes a language (M2), it might include instances of builtins' language entities. In this case, builtins MUST be listed as used language like any other language.{fn-org153}”

ftomassetti:
Also https://github.com/LionWeb-io/specification/issues/367 may be relevant for this issue

ftomassetti:
Also https://github.com/LionWeb-io/specification/issues/380 is related, I think

dslmeinte:
Related: https://github.com/LionWeb-io/lionweb-integration-testing/issues/30

PR #221, opened 2025-06-24
Change LionCore/M3 JSON manually to turn integration tests green

dslmeinte:
Should fix issue #220


ftomassetti:
Looks good to me. All tests pass and the file was probably copied from somewhere, so changing it manually seems the right thing to do.

Thank you!

PR #230, opened 2025-06-28
Revise how we calculate languages to list as used languages

ftomassetti:
Verify that #220 is working: LionCoreBuiltins was already appearing when needed, the issue was probably only present in a file stored in the repository and not generated.

Consider also https://github.com/LionWeb-io/specification/issues/367

We simplify the logic to calculate used languages in SerializationChunks, just looking for the metapointers appearing in the chunk.

### DIFF
--- a/emf-builtins/plugin.xml
+++ b/emf-builtins/plugin.xml
@@ -10,5 +10,12 @@
       </package>
    </extension>
 
-null
+   <extension point="org.eclipse.emf.ecore.generated_package">
+      <!-- @generated builtins -->
+      <package
+            uri="http://lionweb.io/lionweb-java/emf/core/builtins/2023.1"
+            class="io.lionweb.emf.builtins.BuiltinsPackage"
+            genModel="src/main/model/builtins.genmodel"/>
+   </extension>
+
 </plugin>


emf-builtins/plugin.xml
```diff
@@ -10,12 +10,5 @@
...
 
-   <extension point="org.eclipse.emf.ecore.generated_package">
-      <!-- @generated builtins -->
-      <package
-            uri="http://lionweb.io/lionweb-java/emf/core/builtins/2023.1"
-            class="io.lionweb.emf.builtins.BuiltinsPackage"
-            genModel="src/main/model/builtins.genmodel"/>
-   </extension>
-
+null
```
dslmeinte, line 13: I'm guessing this is a change that's unrelated to the rest of the PR. What's its effect?

ftomassetti:
> Can you explain a bit more about what the essence of this change is? In particular, can you explain – e.g. in comments – what “considering” means here?
> 
> I like the shorter code, which doesn't have this “let's traverse through everything again”.

Yes, in essence before we were considering every element touched during serialization to reason of what kind of dependencies the user would need to properly load the chunk, even if those dependencies were not immediately apparent from the chunk. For example, if there is a node Foo with concept C, the language of concept C is an obvious dependency, but if C implements interface I and I is from another language, that other language is an indirect/implicit dependency. Before we were aiming to uncover those, to simplify the life of the user and because this is how I interpreted the meaning of UsedLanguages (maybe I interpreted it as _Necessary Languages for loading_). But since it was clarified that we should just list the languages mentioned in the metapointers I could simplify the code

PR #231, opened 2025-06-28
Add Language.getAnnotationByName

### LINKED ISSUE #203: Add method Language.getAnnotationByName

ftomassetti:
Fix #203 

This method is similar to the existing getConceptByName, getInterfaceByName, etc.

### DIFF
--- a/core/src/test/java/io/lionweb/language/LanguageTest.java
+++ b/core/src/test/java/io/lionweb/language/LanguageTest.java
@@ -1,6 +1,7 @@
 package io.lionweb.language;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
 
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/lionweb/language/LanguageTest.java
```diff
@@ -20,4 +21,24 @@ public void addDependency() {
     m1.addDependency(m3);
     assertEquals(Arrays.asList(m2, m3), m1.dependsOn());
   }
+
+  @Test
+  public void getAnnotationByName() {
+    Language m1 = new Language("m1");
```
dslmeinte, line 27: Minor thing: why are these languages identified as `m1`, etc., instead of `l1`, etc. (or even `lang1`, etc.)?

dslmeinte (approved):

core/src/test/java/io/lionweb/language/LanguageTest.java
ftomassetti: Right! It is because a long time ago Language was called Metamodel, I think. At the time of the first poc. I will change that.

PR #233, opened 2025-06-28
Mark INamed.getName as Nullable

### LINKED ISSUE #196: Make INamed.getName nullable
This is logically dictated by the fact that some overloads of it are nullable

ftomassetti:
Fix #196

### DIFF
--- a/core/src/main/java/io/lionweb/language/Language.java
+++ b/core/src/main/java/io/lionweb/language/Language.java
@@ -142,29 +142,30 @@ public class Language extends M3Node<Language> implements NamespaceProvider, IKe
         .orElse(null);
   }
 
-  public String getName() {
+  public @Nullable String getName() {
     return this.getPropertyValue("name", String.class);
   }
 
   @Override
-  public String getKey() {
+  public @Nullable String getKey() {
     return this.getPropertyValue("key", String.class);
   }
 
-  @Nullable
-  public String getVersion() {
+  public @Nullable String getVersion() {
     return this.getPropertyValue("version", String.class);
   }
 
-  public @Nullable LanguageEntity getElementByName(String name) {
+  public @Nullable LanguageEntity<?> getElementByName(@Nonnull String name) {
+    Objects.requireNonNull(name);
     return getElements().stream()
-        .filter(element -> element.getName().equals(name))
+        .filter(element -> Objects.equals(element.getName(), name))
         .findFirst()
         .orElse(null);
   }
 
-  public @Nullable PrimitiveType getPrimitiveTypeByName(String name) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/lionweb/language/Language.java
```diff
@@ -142,29 +142,30 @@ public Concept requireConceptByName(String name) {
...
   }
 
-  public String getName() {
+  public @Nullable String getName() {
     return this.getPropertyValue("name", String.class);
   }
 
   @Override
-  public String getKey() {
+  public @Nullable String getKey() {
```
dslmeinte, line 150: This is a language's **key** which definitely shouldn't be `null`. (It's also not the name.)

core/src/main/java/io/lionweb/language/Language.java
```diff
@@ -142,29 +142,30 @@ public Concept requireConceptByName(String name) {
...
     return this.getPropertyValue("key", String.class);
   }
 
-  @Nullable
-  public String getVersion() {
+  public @Nullable String getVersion() {
```
dslmeinte, line 154: This is a language's **version** which definitely shouldn't be `null`. (It's also not the name.)

dslmeinte (changes requested):

core/src/main/java/io/lionweb/language/Language.java
ftomassetti: I agree that it _should not_ be null, however we decided to allow the representation of things in a broken state and so all sort of things could have a value that is not valid, I think

ftomassetti:
Rebased to remove the conflict

core/src/main/java/io/lionweb/language/Language.java
enikao: In C#, we solve this by having different implementations of Node -- the default one strongly enforces constraints, and another one is very lose about it.

ftomassetti: Then I would say that:
* The interface `Node`, being the most generic, should the loosest one
* We could introduce another interface `ValidatedNode` that would be stricter. `ValidatedNode` should extend Node, if possible (I am not sure that we can apply restrictions on the parameters from @Nullable to @Nonnull, I would need to verify that)

Would this work? If so, I would create an issue for the introduction of `ValidatedNode` (or `ValidNode`? or `StrictNode`?)

ftomassetti:
Based on the conversation we had this seems a bigger issue which may impact Node but also other classes and interfaces such as INamed, Language, Concept, etc. so https://github.com/LionWeb-io/lionweb-java/issues/238 has been created, while for this PR we will keep the current design to keep things loose

PR #242, opened 2025-07-14
#241 uniformly use `getID()` to access node IDs

### LINKED ISSUE #241: DynamicNode id field and getID method
We discovered the following during the analysis of a performance bottleneck.

In _DynamicNode_, the _equals_ method calls _shallowClassifierInstanceEquality_ which among other things compares the parents using the _getID_ method. However, the same _equals_ method uses the _id_ field, and the _hashCode_ methods is based on the _id_ field, too.

While the _DynamicNode.getID()_ method returns the _id_ field, it's not `final` and can be overridden. In particular, lionweb-kotlin's _BaseNode_ overrides _getID_ but doesn't do anything with the _id_ field.

As a consequence, it can happen that _getID_ and _id_ result in different values. In our specific case, the hash code was always calculated on a null _id_ field, degrading performance significantly when putting many nodes into a set (because it basically becomes a list where on insertion the new element is compared with all the existing ones using _equals_).

I think that either _getID_ should be final, or _id_ should be just an implementation detail (private), with _getID_ and _setID_ as the only ways to access the id. The first choice (final _getID_) would force lionweb-kotlin to adopt a different solution. The second one is IMHO a better design that allows for extensibility while maintaining coherence.

alessiostalla:
This fixes #241

### DIFF
--- a/core/src/main/java/io/lionweb/model/impl/DynamicAnnotationInstance.java
+++ b/core/src/main/java/io/lionweb/model/impl/DynamicAnnotationInstance.java
@@ -36,7 +36,7 @@ public class DynamicAnnotationInstance extends DynamicClassifierInstance<Annotat
     }
     DynamicAnnotationInstance that = (DynamicAnnotationInstance) o;
     return Objects.equals(annotation, that.annotation)
-        && Objects.equals(id, that.id)
+        && Objects.equals(getID(), that.getID())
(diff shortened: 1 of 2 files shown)


ftomassetti:
It looks good and all tests are passing. Thank you @alessiostalla !

PR #243, opened 2025-07-17
Reducing memory occupation and serialization size

ftomassetti:
These improvements were triggered by working on processing a 2M lines of code which leads to ~10M nodes, which leads to ~100M MetaPointers. In that particular system we are using the InMemoryServer so that we actually keep in memory a lot of SerializedClassifierInstance and so we persist a lot of MetaPointers. However even if we did not keep them, we would have to create them when serializing (so yes, they would eventually garbage collected, but still there would be work on allocation and garbage collection).

Based on this I propose preventing duplication of duplicate MetaPointers. We then do something similar for SerializedPropertyValue with a caveat: here we want to re-use common values but not all values. If someone store a huge content in a property then it is probably it will not be duplicate and so we are ok not caching it. However common values (null, false, true, 0, 1, etc.) may be reused a ton of times so we aim to reuse them.

Similarly for serialization we did not take advantage of the possibility of not serializing empty containments and references. We do so to save space during serialization but also to reduce the work during serialization and deserialization.

The first experiment seems encouraging as a particular application significantly reduced the memory used with this PR (by at least 1GB of memory). 

Also, running the SerializationExperiment present in this code base before the PR was getting:
```
= JSON serialization (without compression) =
  serialized in 3013ms
  size 490266650 bytes
```

With this PR we are getting:
```
= JSON serialization (without compression) =
  serialized in 686ms
  size 87614934 bytes
```

### DIFF
--- a/core/src/main/java/io/lionweb/serialization/data/MetaPointer.java
+++ b/core/src/main/java/io/lionweb/serialization/data/MetaPointer.java
@@ -9,21 +9,33 @@ import java.util.*;
 /**
  * A MetaPointer is the combination of the pair Language and Version with a Key, which identify one
  * element within that language.
+ *
+ * <p>We should never have multiple instances with the same value, so equality and identity
+ * coincides for this class.
  */
 public class MetaPointer {
-  private static Map<String, MetaPointer> INSTANCES = new HashMap<>();
+  private static final Map<String, Set<MetaPointer>> INSTANCES_BY_KEY = new HashMap<>();
 
+  /** Provide a MetaPointer with the given value, avoid allocations if unnecessary. */
   public static MetaPointer get(String language, String version, String key) {
-    String hashKey = language + ":" + version + ":" + key;
-    if (!INSTANCES.containsKey(hashKey)) {
-      INSTANCES.put(hashKey, new MetaPointer(language, version, key));
+    Set<MetaPointer> entryForKey = INSTANCES_BY_KEY.computeIfAbsent(key, k -> new HashSet<>());
+    Optional<MetaPointer> match =
+        entryForKey.stream()
+            .filter(
+                mp -> Objects.equals(mp.language, language) && Objects.equals(mp.version, version))
+            .findFirst();
+    if (match.isPresent()) {
+      return match.get();
+    } else {
+      MetaPointer metaPointer = new MetaPointer(language, version, key);
+      entryForKey.add(metaPointer);
+      return metaPointer;
     }
-    return INSTANCES.get(hashKey);
   }
 
-  private String key;
-  private String version;
-  private String language;
+  private final String key;
+  private final String version;
+  private final String language;
 
   private MetaPointer(String language, String version, String key) {
     this.key = key;
@@ -64,17 +76,12 @@ public class MetaPointer {
 
   @Override
   public boolean equals(Object o) {
-    if (this == o) return true;
-    if (!(o instanceof MetaPointer)) return false;
-    MetaPointer that = (MetaPointer) o;
-    return Objects.equals(key, that.key)
-        && Objects.equals(version, that.version)
-        && Objects.equals(language, that.language);
+    return this == o;
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(key, version, language);
+    return System.identityHashCode(this);
   }
 
   @Override
--- a/core/src/main/java/io/lionweb/serialization/data/MetaPointer.java
+++ b/core/src/main/java/io/lionweb/serialization/data/MetaPointer.java
@@ -1,5 +1,7 @@
 package io.lionweb.serialization.data;
 
+import com.google.common.collect.Multimap;
+import com.google.common.collect.Multimaps;
 import io.lionweb.language.Feature;
 import io.lionweb.language.IKeyed;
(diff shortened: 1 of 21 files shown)


core/src/main/java/io/lionweb/serialization/data/MetaPointer.java
```diff
@@ -4,90 +4,84 @@
...
  * element within that language.
+ *
+ * <p>We should never have multiple instances with the same value, so equality and identity
+ * coincides for this class.
  */
 public class MetaPointer {
-  private String key;
-  private String version;
-  private String language;
+  private static final Map<String, Set<MetaPointer>> INSTANCES_BY_KEY = new HashMap<>();
```
dslmeinte, line 17: Why not use a `Map<String, Map<String, Map<String, MetaPointer>>>` so you could do a look up that's essentially `INSTANCES.getOrDefault(language, Map.EMPTY)?.getOrDefault(version, Map.EMPTY)?.getOrDefault(key,  null)` (or something similar that's equivalent, e.g. using `Map.computeIfAbsent(...)`, but might be performing better)?

core/src/main/java/io/lionweb/serialization/data/SerializedPropertyValue.java
```diff
@@ -4,34 +4,29 @@
 import java.util.Map;
 import java.util.Objects;
 
-/** This represents the serialization of the value of a property in a Node. */
+/**
+ * This represents the serialization of the value of a property in a Node. This class is immutable
+ * and multiple coinciding value share the same identity, but this is not guaranteed for all
+ * instances.
+ */
 public class SerializedPropertyValue {
```
dslmeinte, line 12: Maybe add a remark to the implementation of `hashCode()` why you can't use `System.identityHashCode()` here?

dslmeinte (approved):

core/src/main/java/io/lionweb/serialization/data/MetaPointer.java
enikao, line 17: This will prevent the garbage collector to claim unused `MetaPointer` instances.

The concept we want here is "interning", Java does it a lot for [Strings](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#intern--)

A quick search turns up articles like https://www.javacodegeeks.com/2014/03/java-object-interning.html that generalizes to arbitrary objects.

enikao: at least the original implementation very much looks like a multimap, readily available in Guava.

ftomassetti: I have quickly look to interning and also to the flag for deduplicating strings (which helped a bit).
I see the problem with the garbage collection never happening but I am more worried for that in relation to SerializedPropertyValue then for Metapointers.

I will look into interning.

We could also consider keeping only the most recently used values up to a certain size (4096 entries)? But then we would lose the guarantee identity = equality

ftomassetti: I tried the two approaches and run the SerializationExperiment 3 times:

```
MULTIMAP

  private static final Multimap<String, MetaPointer> INSTANCES_BY_KEY =
      Multimaps.newSetMultimap(new WeakHashMap<>(), HashSet::new);

  /** Provide a MetaPointer with the given value, avoid allocations if unnecessary. */
  public static MetaPointer get(String language, String version, String key) {
    Collection<MetaPointer> entryForKey = INSTANCES_BY_KEY.get(key);
    Optional<MetaPointer> match =
        entryForKey.stream()
            .filter(
                mp -> Objects.equals(mp.language, language) && Objects.equals(mp.version, version))
            .findFirst();
    if (match.isPresent()) {
      return match.get();
    } else {
      MetaPointer metaPointer = new MetaPointer(language, version, key);
      entryForKey.add(metaPointer);
      return metaPointer;
    }
  }

Tree generated in 7937-8043 ms
= JSON serialization (without compression) =
  serialized in 508-607ms
  size 87614934 bytes
  unserialized in 1883-2001ms

MAPS OF MAP

  private static final Map<String, Map<String, WeakHashMap<String, MetaPointer>>> INSTANCES_BY_KEY_LANGUAGE_VERSION =
       new HashMap<>();

  /** Provide a MetaPointer with the given value, avoid allocations if unnecessary. */
  public static MetaPointer get(String language, String version, String key) {
    return INSTANCES_BY_KEY_LANGUAGE_VERSION.computeIfAbsent(key, k -> new HashMap<>())
            .computeIfAbsent(language, l -> new WeakHashMap<>())
            .computeIfAbsent(version, v -> new MetaPointer(language, version, key));
  }

Tree generated in 7721-8264 ms
= JSON serialization (without compression) =
  serialized in 590-688ms
  size 87614934 bytes
  unserialized in 1671-1817ms
```

They seem very close in serialization, while there is an advantage in deserialization with the maps of maps, so I would go for it. However I would make the most internal map a WeakHashMap, not to prevent garbage collection

ftomassetti: I went for using `WeakHashMap`, so that garbage collection is not prevented by this map

PR #246, opened 2025-08-01
Bugfix in id assignation for InMemoryServer

ftomassetti:
In an InMemoryServer we could load some nodes and then let the user create new nodes. Now, if we load nodes which happens to have IDs using the same format as the ids we generate there is the risk of a conflict.

So when assigning IDs to users we check they have not been already used.

### DIFF
--- a/client/src/main/java/io/lionweb/client/inmemory/RepositoryData.java
+++ b/client/src/main/java/io/lionweb/client/inmemory/RepositoryData.java
@@ -82,8 +82,11 @@ class RepositoryData {
 
   List<String> ids(int count) {
     List<String> res = new ArrayList<>(count);
-    for (int i = 0; i < count; i++) {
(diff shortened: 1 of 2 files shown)


dslmeinte (approved):
The additional check can be quite (performance-)costly, though!

ftomassetti:
> The additional check can be quite (performance-)costly, though!

Thank you @dslmeinte ! I was hoping that the check into the map was quick, but I will keep an eye on this and see if I notice a performance degradation

dslmeinte:
> > The additional check can be quite (performance-)costly, though!
> 
> Thank you @dslmeinte ! I was hoping that the check into the map was quick, but I will keep an eye on this and see if I notice a performance degradation

The check in the map is indeed quick, but what happens if you add loads of nodes through the API with IDs in that format?

PR #248, opened 2025-08-17
Adding possibility to observe nodes

ftomassetti:
Here we provide the code to observe nodes. This is a pre-requisite to implement Delta and the code has been extracted from the branch where delta is being implemented. 

- [x] If a Node that was root becomes not root, then its partitions observers should be removed
- [x] When a node is added into a tree it should update its observer (perhaps we could add a flag called `dirty` to make it knows that it _should_ update the cache at the next access)

Given this change was very impactful we started measuring coverage, and set the rule to get coverage to 60%+ for each file tested so we added a lot of tests and comments as needed.

### DIFF
--- a/core/src/main/java/io/lionweb/model/impl/DynamicClassifierInstance.java
+++ b/core/src/main/java/io/lionweb/model/impl/DynamicClassifierInstance.java
@@ -104,6 +104,9 @@ public abstract class DynamicClassifierInstance<T extends Classifier<T>>
   public void addChild(@Nonnull Containment containment, @Nonnull Node child) {
     Objects.requireNonNull(containment);
     Objects.requireNonNull(child);
+    if (!getClassifier().allContainments().contains(containment)) {
+      throw new IllegalArgumentException();
+    }
     if (containment.isMultiple()) {
       addContainment(containment, child);
     } else {
--- a/core/src/main/java/io/lionweb/model/impl/DynamicNode.java
+++ b/core/src/main/java/io/lionweb/model/impl/DynamicNode.java
@@ -181,14 +181,15 @@ public class DynamicNode extends DynamicClassifierInstance<Concept>
 
     String containmentValueStr = "<null>";
     if (containmentValues != null) {
-      containmentValues.entrySet().stream()
-          .map(
-              e -> {
-                String childrenRepr =
-                    e.getValue().stream().map(c -> c.getID()).collect(Collectors.joining(", "));
-                return e.getKey() + "=" + childrenRepr;
-              })
-          .collect(Collectors.joining(", "));
+      containmentValueStr =
+          containmentValues.entrySet().stream()
+              .map(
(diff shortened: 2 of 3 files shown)


core/src/main/java/io/lionweb/model/impl/AbstractClassifierInstance.java
```diff
@@ -44,35 +44,47 @@ public abstract class AbstractClassifierInstance<T extends Classifier<T>>
...
     if (instance instanceof DynamicAnnotationInstance) {
       ((DynamicAnnotationInstance) instance).setAnnotated(null);
     }
+    if (observer != null) {
+      observer.annotationRemoved(this, index, instance);
+    }
+    return index;
   }
 
   void tryToRemoveAnnotation(@Nonnull AnnotationInstance instance) {
```
enikao, line 90: Why does this not return an `int`, or `bool`, or maybe `Integer` (with `null` meaning "not removed")?

enikao (commented):
In delta protocol, we consciously decided to only have observers on _partitions_.
Main reason: If they're on nodes, each observer (or a shared superclass) has to keep track of all newly added / removed nodes, and keep attaching / removing itself from them. It also gets more tricky if we moved a node from on parent to the other, and have observers on both -- who gets the message?

Implementation-wise this very simple in C#: We walk up parents until we find the root (or first `IPartitionInstance`, we have that there). If we find one, we check if there's an observer registered, and do the same thing as here.

Of course, this implementation can be used to satisfy delta spec. But I think we should only have two pretty different strategies for such a similar task if we have a good reason to do so.

ftomassetti:
> Before doing any optimization on this I'd measure the actual impact.

Normally I would postpone optimizations, but for years  I have been fighting a lot of performance issues and at this time we are using LionWeb Java for projects where we cannot afford worsening performance. For this reason so I want to be careful and ensure we do not merge anything that would make performance significantly worse when observers are not used. If a performance hit is acceptable for those who needs to use observers, others should not experience any change.

The problem with measuring the impact would be that it is easy to do "in the wild", using the library for a while on different use-cases, but it is for me difficult to reproduce it "in lab", for this reason I tend to err on the side of caution.

> If we see some, we could store an internal reference (or null) from each node to the observer -- without having an individual node observer API.

So that would mean having a field with the pointer to the observer, with the only difference that the public methods to register/unregister observers would work only for partitions, right?

> We could even store only a boolen if there's any observer.

That also would be reasonable

core/src/main/java/io/lionweb/model/impl/AbstractClassifierInstance.java
ftomassetti: (the comment is on the interface from which it is overridden)

ftomassetti:
Applied the change to make sure that observers can be added only to partitions, and added a good amount of tests.

PR #249, opened 2025-08-29
Client propagation

ftomassetti:
In LionWebClient we were not forwarding a couple of calls to the underlying `chunkLevelBulkAPIs` element.

Also, added a few methods in `SerializedClassifierInstance` and `SerializedContainmentValue`, with some tests.

### DIFF
--- a/client/src/main/java/io/lionweb/client/LionWebClient.java
+++ b/client/src/main/java/io/lionweb/client/LionWebClient.java
@@ -276,15 +276,15 @@ public class LionWebClient
   @Override
   public RepositoryVersionToken createPartitionsFromChunk(
       @NotNull List<SerializedClassifierInstance> data) throws IOException {
-    throw new UnsupportedOperationException();
+    return this.chunkLevelBulkAPIs.createPartitionsFromChunk(data);
   }
 
-  @Nullable
-  @Override
-  public RepositoryVersionToken storeChunk(@NotNull List<SerializedClassifierInstance> nodes)
-      throws IOException {
-    throw new UnsupportedOperationException();
-  }
+    @Nullable
+    @Override
+    public RepositoryVersionToken storeChunk(@NotNull List<SerializedClassifierInstance> nodes)
+            throws IOException {
+        return this.chunkLevelBulkAPIs.storeChunk(nodes);
+    }
 
   @NotNull
   @Override
--- a/client/src/main/java/io/lionweb/client/LionWebClient.java
+++ b/client/src/main/java/io/lionweb/client/LionWebClient.java
@@ -290,14 +290,14 @@ public class LionWebClient
   @Override
   public List<SerializedClassifierInstance> retrieveAsChunk(
       @Nullable List<String> nodeIds, int limit) throws IOException {
-    throw new UnsupportedOperationException();
+    return this.chunkLevelBulkAPIs.retrieveAsChunk(nodeIds, limit);
   }
 
   @NotNull
   @Override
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/lionweb/serialization/data/SerializedClassifierInstance.java
```diff
@@ -169,6 +169,25 @@ public void addChildren(MetaPointer containment, List<String> childrenIds) {
     this.containments.add(new SerializedContainmentValue(containment, childrenIds));
   }
 
+  /**
+   * Removes the specified child from the containments, if present.
+   *
+   * @param child the identifier of the child to be removed; must not be null
+   * @return true if the child was successfully removed, false otherwise
+   */
+  public boolean removeChild(@Nonnull String child) {
```
dslmeinte, line 178: We're talking about the child's **ID** here, right? I'd prefer if the argument‘s name reflected that.

core/src/main/java/io/lionweb/serialization/data/SerializedContainmentValue.java
```diff
@@ -30,6 +31,18 @@ public void setValue(List<String> value) {
...
   }
 
+  /**
+   * Removes a child from the list of Node-IDs contained.
+   *
+   * @param child the identifier of the child to be removed; must not be null
+   * @return true if the child was successfully removed, false otherwise
+   * @throws NullPointerException if the child is null
+   */
+  public boolean removeChild(@Nonnull String child) {
```
dslmeinte, line 41: Same here: `childID` would be clearer.

core/src/test/java/io/lionweb/serialization/data/SerializedClassifierInstanceTest.java
```diff
@@ -0,0 +1,231 @@
...
+    assertNull(sci.getClassifier());
+    assertNull(sci.getParentNodeID());
+
+    assertTrue(sci.getProperties().isEmpty());
+    assertTrue(sci.getContainments().isEmpty());
+    assertTrue(sci.getChildren().isEmpty());
+    assertTrue(sci.getReferences().isEmpty());
+    assertTrue(sci.getAnnotations().isEmpty());
+
+    assertThrows(UnsupportedOperationException.class, () -> sci.getProperties().add(null));
```
dslmeinte, line 30: Why would adding a `null` property cause a `UnsupportedOperationException` to be thrown? I'd expect a less generic, more precise exception.

client/src/main/java/io/lionweb/client/LionWebClient.java
```diff
@@ -276,28 +276,28 @@ public LionWebVersion getLionWebVersion() {
   @Override
   public RepositoryVersionToken createPartitionsFromChunk(
       @NotNull List<SerializedClassifierInstance> data) throws IOException {
-    throw new UnsupportedOperationException();
+    return this.chunkLevelBulkAPIs.createPartitionsFromChunk(data);
```
dslmeinte, line 279: These methods (of `ClientForChunkLevelBulkAPIs`) are now hooked up, but I don't see them being unit tested.

core/src/test/java/io/lionweb/serialization/data/SerializedClassifierInstanceTest.java
ftomassetti: That would be the case for each value (I will change the value). The problem is that getProperties return immutable lists, so try to add to them cause this exception to be thrown. The exception is determined by the implementation of the immutable list

dslmeinte (approved):

ftomassetti:
Thank you @dslmeinte !

PR #255, opened 2025-09-06
Improving binary format

ftomassetti:
After comparing flatbuffers and protobuf, protobuf seems to be overall better, so we focused on refining the structure used, interning also languages and checking which values are optional.

Also:
* Updated protobuf version
* Used the versions catalog more consistently
* Renamed `UsedLanguage` into `LanguageVersion` and interning it

# Assessment

I run the [experiment code for serialization and deserialization stored in this repository](https://github.com/LionWeb-io/lionweb-java/blob/main/core/src/experiments/java/io/lionweb/experiments/SerializationAndDeserializationExperiment.java), and with the code in main I got:

```
= ProtoBuf serialization =
serialized in 391ms 
size 6,017,530 bytes
unserialized in 2745ms
```

With these changes I got:
```
= ProtoBuf serialization =
serialized in 355ms (and then 270ms) 
size 6,018,793 bytes
unserialized in 1573ms (and then 1419ms)
```

So unserialization seems notably faster, while serialization is a little bit faster. Surprisingly size increased slightly (however the same file in JSON takes 87,614,934 and in flatbuffers 16,777,216, so it is much smaller anyway).

I also ran an experiment with data used in a project (2806 partitions). There with the code in main I got:

```
Total size unzipped: 505,526,989
Total size zipped: 126,770,843
Max partition size: 22,890,945
Serialized to PROTOBUF in 6674 ms
Unserialized 2806 chunks in 6801 ms
```

With these changes I got:
```
Total size unzipped: 504,699,469
Total size zipped: 126,654,215
Max partition size: 22,893,581
Serialized to PROTOBUF in 5207 ms (and then 4522ms)
Unserialized 2806 chunks in 6694 ms (and then 6709ms)
```

So the size seems overall very slightly less (however not in the case of the largest partition). Serialization seems significantly better (68% ~~78%~~ of the original time) and unserialization slightly better (98%), but within the margins of error of this experiment.

# Note on functional tests

Functional tests now refers to an unreleased version of the LionWeb Server, using code from a non merged branch.
If this is approved, I will open the PR for the LionWeb Server, which is ready.

### DIFF
--- a/core/src/main/java/io/lionweb/serialization/ProtoBufSerialization.java
+++ b/core/src/main/java/io/lionweb/serialization/ProtoBufSerialization.java
@@ -115,14 +115,14 @@ public class ProtoBufSerialization extends AbstractSerialization {
               n.getContainmentsList()
                   .forEach(
                       c -> {
-                        if (c.getSiChildrenList().stream().anyMatch(el -> el == 0)) {
-                          throw new DeserializationException(
-                              "Unable to deserialize child identified by Null ID");
+                        List<String> children = new ArrayList<>(c.getSiChildrenList().size());
+                        for (int childIndex : c.getSiChildrenList()) {
+                          if (childIndex == 0) {
+                            throw new DeserializationException(
+                                "Unable to deserialize child identified by Null ID");
+                          }
+                          children.add(stringsArray[childIndex]);
                         }
-                        List<String> children =
-                            c.getSiChildrenList().stream()
-                                .map(el -> stringsArray[el])
-                                .collect(Collectors.toList());
                         if (!children.isEmpty()) {
                           SerializedContainmentValue scv =
                               new SerializedContainmentValue(


core/src/main/proto/io/lionweb/protobuf/Chunk.proto
```diff
@@ -7,53 +7,58 @@ option java_package = "io.lionweb.protobuf";
...
+  // Yes, we should _not_ have null IDs, but if necessary we can represent them
+  optional uint32 id = 1;
+  uint32 classifier = 2;
   repeated PBProperty properties = 3;
   repeated PBContainment containments = 4;
   repeated PBReference references = 5;
   /* This is a list of indexes representing the string values corresponding to the IDs of the annotations */
-  repeated int32 annotations = 6;
-  /* Optional*/ int32 parent = 7;
+  /* int to accomodate potential null values */
```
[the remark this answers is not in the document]
ftomassetti: We used `int32` in two situations: for list of annotations and list of children. This allows to represent the _incorrect_ situation in which we have `null` listed as a child or an annotation. Correct chunks would not have any of these values to null and therefore no negative value in these lists. 

My understanding is that this would impact only incorrect chunks and for those cases they would get 3 extra bytes for each of these `null` values.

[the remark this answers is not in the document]
enikao: I guess this comment can go then?

core/src/main/proto/io/lionweb/protobuf/Chunk.proto
```diff
@@ -6,54 +6,71 @@ option java_multiple_files = true;
...
+  optional uint32 si_id = 1;
+  uint32 mpi_classifier = 2;
   repeated PBProperty properties = 3;
   repeated PBContainment containments = 4;
   repeated PBReference references = 5;
   /* This is a list of indexes representing the string values corresponding to the IDs of the annotations */
-  repeated int32 annotations = 6;
-  /* Optional*/ int32 parent = 7;
+  /* int to accomodate potential null values */
+  repeated uint32 si_annotations = 6 [packed = true];
```
enikao, line 47: I think in [proto3](https://protobuf.dev/programming-guides/proto3/#use-packed) `packed=true` is the default, so we can omit it?

core/src/main/proto/io/lionweb/protobuf/Chunk.proto
```diff
@@ -6,54 +6,71 @@ option java_multiple_files = true;
...
-  int32 key = 1;
-  int32 version = 2;
+  optional uint32 si_key = 1;
+  optional uint32 si_version = 2;
 }
 
 message PBNode {
-  int32 id = 1;
-  int32 classifier = 2;
+  // Yes, we should _not_ have null IDs, but if necessary we can represent them
```
enikao, line 39: Now we have two ways of encoding `null` -- not having this field, and having this field with value `0`.

core/src/main/proto/io/lionweb/protobuf/Chunk.proto
ftomassetti: Yes, it is true, however it may be not obvious, so I would prefer to have it and I added it also below for children

core/src/main/proto/io/lionweb/protobuf/Chunk.proto
enikao: Which are these situations? I think `uint32` [defaults](https://protobuf.dev/programming-guides/proto3/#default) to `0`, and Protobuf doesn't actually serialize the default value.

core/src/test/java/io/lionweb/serialization/simplemath/Sum.java
```diff
@@ -59,4 +60,15 @@ protected List<? extends Node> concreteGetChildren(Containment containment) {
     }
     return super.concreteGetChildren(containment);
   }
+
+  @Override
+  public void addChild(@Nonnull Containment containment, @Nonnull Node child) {
```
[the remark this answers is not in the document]
ftomassetti: Yes, absolutely. But this is a simple class used in tests and I want to make it obvious

core/src/main/proto/io/lionweb/protobuf/Chunk.proto
enikao: From this (same link as above):

> Note that for implicit-presence **scalar** fields, once a message is parsed there’s no way of telling whether that field was explicitly set to the default value (for example whether a boolean was set to false) or just not set at all: you should bear this in mind when defining your message types. For example, don’t have a boolean that switches on some behavior when set to false if you don’t want that behavior to also happen by default. Also note that **if a scalar message field is set to its default, the value will not be serialized on the wire.** If a float or double value is set to +0 it will not be serialized, but -0 is considered distinct and will be serialized.

I interpret that a _required_ `uint32` field with value `0` will not be serialized.

ftomassetti: I interpret this differently as it says **implicit** and in the documentation says that:
> there are two types of singular fields
It then list **implicit** and **optional**.

So my understanding is that an *optional* field is **not** *implicit* and therefore the paragraph you reported does not apply.

If I look at the generated code for optional field I can see this:

```
    public Builder setSiId(int value) {

      siId_ = value;
      bitField0_ |= 0x00000001;
      onChanged();
      return this;
    }
```

It seems that the fact the value is set is tracked separately in the `bitfield`.

It can also being controlled to an explicit clear operation:

```
    public Builder clearSiId() {
      bitField0_ = (bitField0_ & ~0x00000001);
      siId_ = 0;
      onChanged();
      return this;
    }
```

It can also be inspected:

```
public boolean hasSiId() {
    return ((bitField0_ & 0x00000001) != 0);
  }
```

And the bitField controls if the value is actually written at all:
```
@java.lang.Override
  public void writeTo(com.google.protobuf.CodedOutputStream output)
                      throws java.io.IOException {
    getSerializedSize();
    if (((bitField0_ & 0x00000001) != 0)) {
      output.writeUInt32(1, siId_);
    }
    ...omitted
  }
 ```

**Review by dslmeinte — APPROVED** [review]

File: core/src/main/proto/io/lionweb/protobuf/Chunk.proto (hunk 3 above)
Comment (enikao) [reply core/src/main/proto/io/lionweb/protobuf/Chunk.proto:39]: at least in C# it works:

```protobuf
message PsTest {
  optional uint32 opt = 1;
  uint32 req = 2;
}
```

```csharp
    [TestMethod]
    public void Test()
    {
        Print("empty", new PsTest());
        Print("req = 0", new PsTest() { Req = 0 });
        Print("opt = 0", new PsTest() { Opt = 0 });
        Print("req = 0, opt = 0", new PsTest() { Req = 0, Opt = 0 });
        Print("req = 1", new PsTest() { Req = 1 });
        Print("opt = 1", new PsTest() { Opt = 1 });
        Print("req = 1, opt = 1", new PsTest() { Req = 1, Opt = 1 });
    }

    private PsTest Print(string name, PsTest psTest)
    {
        var buf = psTest.ToByteArray();
        var bufContents = string.Join("", buf.Where(b => b != 0).Select(b => b.ToString("x2")));
        Console.WriteLine($"{name}:\n  toString: {psTest}\n  buf[{buf.Length}]: 0x{bufContents}");

        return psTest;
    }
```

printout:
```
empty:
  toString: { }
  buf[0]: 0x
req = 0:
  toString: { }
  buf[0]: 0x
opt = 0:
  toString: { "opt": 0 }
  buf[2]: 0x08
req = 0, opt = 0:
  toString: { "opt": 0 }
  buf[2]: 0x08
req = 1:
  toString: { "req": 1 }
  buf[2]: 0x1001
opt = 1:
  toString: { "opt": 1 }
  buf[2]: 0x0801
req = 1, opt = 1:
  toString: { "opt": 1, "req": 1 }
  buf[4]: 0x08011001
```

PR #264, opened 2025-09-26
SerializationChunk and duplicate features

### LINKED ISSUE #262: Serialization Chunk Validation - Double features
We should verify that no feature is present twice

### LINKED ISSUE #263: Fix SerializedClassifierInstance.setPropertyValue
```
  public void setPropertyValue(MetaPointer property, String serializedValue) {
    this.properties.add(SerializedPropertyValue.get(property, serializedValue));
  }
```

If the entry is already present, it should be replaces

ftomassetti:
We make a safe method to add a property value to a SerializedClassifierInstance avoid duplication (fix #263 )

We add a chunk for validating chunks against duplicated features (fix #262 )

### DIFF
--- a/core/src/test/java/io/lionweb/utils/ChunkValidatorTest.java
+++ b/core/src/test/java/io/lionweb/utils/ChunkValidatorTest.java
@@ -242,6 +242,208 @@ public class ChunkValidatorTest {
   }
 
   @Test
+  public void testDuplicateContainmentMetaPointers() {
+    SerializedClassifierInstance parent = createValidNode("parent1", null);
+    SerializedClassifierInstance child1 = createValidNode("child1", "parent1");
+    SerializedClassifierInstance child2 = createValidNode("child2", "parent1");
+
+    // Create two containments with the same metapointer
+    SerializedContainmentValue containment1 =
+        new SerializedContainmentValue(testMetaPointer, Arrays.asList("child1"));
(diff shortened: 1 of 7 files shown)


core/src/test/java/io/lionweb/utils/ChunkValidatorTest.java
```diff
@@ -241,6 +241,208 @@ public void testNodeWithPropertiesAndReferences() {
...
+    SerializedClassifierInstance child1 = createValidNode("child1", "parent1");
+    SerializedClassifierInstance child2 = createValidNode("child2", "parent1");
+
+    // Create two containments with the same metapointer
+    SerializedContainmentValue containment1 =
+        new SerializedContainmentValue(testMetaPointer, Arrays.asList("child1"));
+    SerializedContainmentValue containment2 =
+        new SerializedContainmentValue(testMetaPointer, Arrays.asList("child2"));
+
+    parent.addContainmentValue(containment1);
```
dslmeinte, line 256: Shouldn't this be `parent.unsafeAddContainmentValue`? I'd that method to exist, just as `unsafeAddPropertyValue`.

ftomassetti: You are right. I reviewed other similar methods to be renamed with the `unsafe` prefix

PR #272, opened 2025-10-24
Initial Delta Support

ftomassetti:
With this PR we setup the infrastructure to support the delta protocol.

* We create the basic classes for commands and events (moved to #276)
* We define the main interfaces
* We add the possibility to monitor nodes for changes, considering only property changes for now
* We add support in the InMemoryServer to execute commands for property changes

This is intended to collect early feedback before implementing support for other commands and events.

Only then we will implement the DeltaChannel that actually exchanges messages through WebSockets.

In preparation for this we extracted the PRs #274 #275 #276 and #277

### DIFF
--- a/client/src/main/java/io/lionweb/client/delta/DeltaClient.java
+++ b/client/src/main/java/io/lionweb/client/delta/DeltaClient.java
@@ -33,8 +33,8 @@ public class DeltaClient implements DeltaEventReceiver, DeltaQueryResponseReceiv
   private MonitoringObserver observer = new MonitoringObserver();
   private String participationId;
   private HashMap<String, Set<WeakReference<ClassifierInstance<?>>>> nodes = new HashMap<>();
-  private PrimitiveValuesSerialization primitiveValuesSerialization =
-      new PrimitiveValuesSerialization();
+  private DataTypesValuesSerialization dataTypesValuesSerialization =
+      new DataTypesValuesSerialization();
   private AbstractSerialization serialization;
   private Set<String> queriesSent = new HashSet<>();
   private String clientId;
@@ -49,7 +49,7 @@ public class DeltaClient implements DeltaEventReceiver, DeltaQueryResponseReceiv
     this.channel = channel;
     this.channel.registerEventReceiver(this);
     this.channel.registerQueryResponseReceiver(this);
-    this.primitiveValuesSerialization.registerLionBuiltinsPrimitiveSerializersAndDeserializers(
+    this.dataTypesValuesSerialization.registerLionBuiltinsPrimitiveSerializersAndDeserializers(
         lionWebVersion);
     this.serialization = SerializationProvider.getStandardJsonSerialization(lionWebVersion);
     this.serialization.setUnavailableParentPolicy(UnavailableNodePolicy.PROXY_NODES);
@@ -201,7 +201,7 @@ public class DeltaClient implements DeltaEventReceiver, DeltaQueryResponseReceiv
                   commandId,
                   classifierInstance.getID(),
                   MetaPointer.from(property),
-                  primitiveValuesSerialization.serialize(property.getType().getID(), newValue)));
+                  dataTypesValuesSerialization.serialize(property.getType().getID(), newValue)));
     }
 
     @Override


client/src/main/java/io/lionweb/client/delta/messages/DeltaQueryResponse.java
```diff
@@ -0,0 +1,17 @@
+package io.lionweb.client.delta.messages;
+
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import org.jetbrains.annotations.NotNull;
+
+public class DeltaQueryResponse {
```
enikao: I think we need to specialize this for every query, as they return different information.

Example: `SignOn` returns participationId, but `ListPartitions` returns a list of partitions.

client/src/main/java/io/lionweb/client/delta/InMemoryDeltaChannel.java
```diff
@@ -0,0 +1,53 @@
+package io.lionweb.client.delta;
+
+import io.lionweb.client.delta.messages.*;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.function.Function;
+import org.jetbrains.annotations.Nullable;
+
+public class InMemoryDeltaChannel implements DeltaChannel {
```
enikao: When implementing C#, I first thought it would be very nice for debugging to have both the client and the server in the same process. It turned out to be _very_ confusing, as we have several instances of the same node (with the same id) to handle.
To me, it now feels more convenient to have them in separate processes.

YMMV, just sharing some pain (-:

client/src/main/java/io/lionweb/client/inmemory/InMemoryServer.java
```diff
@@ -182,6 +229,53 @@ public Map<String, ClassifierResult> nodesByLanguage(
...
+        RepositoryData repositoryData = getRepository(repositoryName);
+        List<SerializedClassifierInstance> retrieved = new ArrayList<>();
+        try {
+          repositoryData.retrieve(changeProperty.node, 0, retrieved);
+        } catch (IllegalArgumentException e) {
+          return DeltaCommandResponse.rejected(
+              "Node with id " + changeProperty.node + " not found");
+        }
+        SerializedClassifierInstance node = retrieved.get(0);
+        String oldValue = node.getPropertyValue(((ChangeProperty) command).property);
```
enikao: I tried to separate the protocol handling from _applying_ the deltas to a local model, and this turned out to be helpful: In our project, we're actually not using the repository, but recording all deltas on one side, sending them to the other side, and apply them there. We can reuse all the logic written for repository.

client/src/main/java/io/lionweb/client/delta/messages/commands/annotations/AddAnnotation.java
```diff
@@ -0,0 +1,25 @@
...
+import io.lionweb.serialization.data.SerializationChunk;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Add new node newAnnotation to parent's annotations at index. newAnnotation might be a single node
+ * or an arbitrary complex subtree. All nodes in that subtree MUST be new, i.e. their id MUST NOT
+ * exist in the repository. Nodes in that subtree MAY have references to already existing nodes, and
+ * already existing nodes MAY have references to nodes in that subtree.
+ */
+public final class AddAnnotation extends DeltaCommand {
```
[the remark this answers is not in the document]
ftomassetti: For now we are on Java 8, with the prospect to move to Java 11. Records were introduced in Java 14, so they are not available and they will not for the foreseeable future

client/src/main/java/io/lionweb/client/delta/messages/DeltaQueryResponse.java
ftomassetti: Making this class abstract

ftomassetti: I made this class abstract and created concrete subclasses

PR #292, opened 2025-12-24
Refactor `standardInitialization` to support custom LionWeb versions

### LINKED ISSUE #289: Consider making the standardInitialization method public
_standardInitialization_ is used to set up an _AbstractSerialization_ instance with the plumbing necessary to deal with LionWeb built-in languages.

This method is protected, so it can't be used to set up a complex serialization scheme in user code.

Also, the method assumes that you want the same LW version as the serialization format, which is usually the case. But in complex scenarios, when you want to deal with multiple LW versions (e.g. to perform a migration), it's necessary to support multiple versions of the builtins. So, a method overload taking the LW version explicitly instead of getting it out of the _AbstractSerialization_ would be useful to avoid repetition.

ftomassetti:
Overload `standardInitialization` to allow specifying a custom `LionWebVersion` for core languages.

Fix #289

### DIFF
--- a/core/src/main/java/io/lionweb/serialization/SerializationProvider.java
+++ b/core/src/main/java/io/lionweb/serialization/SerializationProvider.java
@@ -69,7 +69,7 @@ public class SerializationProvider {
    * getStandardJsonSerialization instead.
    *
    * <p>This method allows to consider core languages from a different LionWeb Version than the one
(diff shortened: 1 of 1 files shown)


dslmeinte (approved):

alessiostalla (approved):
Thanks!

PR #300, opened 2026-01-19
Store Languages

### LINKED ISSUE #297: Package languages in json format in the jar
We may want to make the gradle plugin also add the languages under src/main/lionweb be packed under resources or a special subdirectories, so that modules using this package as a dependency may found the languages and use them.

Use case:
- In module A I define the languages as JSON files and then I generate the language classes
- In module B, depending on module A, I want to generate the node classes from the same JSON files

### LINKED ISSUE #298: Add flag to control which languages to generate
We may need to put under src/main/lionweb not just the languages from which we want to generate classes but also their dependencies. Currently we generate code for all the languages under src/main/lionweb. It would be useful to be able to control from which languages we generate classes and which ones are just dependencies

ftomassetti:
- We store LionWeb languages in JSON format within the JAR, under META-INF, so that other modules depending on the modules containing the languages, can access them (Fix #297)
- We let the user define which languages to generate, as we may have more languages present just as dependencies (Fix #298)
- We let the user specify mappings, i.e., classes we have written manually or obtained elsewhere to represent certain concepts or other language elements

# TODO
- [x] Document the configuration parameters of the Gradle Plugin

### DIFF
--- a/gradle-plugin/src/main/java/io/lionweb/gradleplugin/tasks/AbstractGenerationTask.java
+++ b/gradle-plugin/src/main/java/io/lionweb/gradleplugin/tasks/AbstractGenerationTask.java
@@ -71,15 +71,25 @@ public abstract class AbstractGenerationTask extends DefaultTask {
                   JarEntry entry = entries.nextElement();
                   if (entry.getName().startsWith("META-INF/lionweb/")
                       && entry.getName().endsWith(".json")) {
-                    InputStream inputStream = jarFile.getInputStream(entry);
-                    JsonElement je =
-                        JsonParser.parseReader(new java.io.InputStreamReader(inputStream));
-                    dependenciesChunks.add(
-                        new LowLevelJsonSerialization().deserializeSerializationBlock(je));
+                    try {
+                      InputStream inputStream = jarFile.getInputStream(entry);
+                      JsonElement je =
+                          JsonParser.parseReader(new java.io.InputStreamReader(inputStream));
+                      dependenciesChunks.add(
+                          new LowLevelJsonSerialization().deserializeSerializationBlock(je));
+                    } catch (Exception e) {
+                      getLogger()
+                          .error(
+                              "Error reading jar file entry: "
+                                  + jar.getAbsolutePath()
+                                  + "#"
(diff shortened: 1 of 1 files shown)


gradle-plugin/src/main/java/io/lionweb/gradleplugin/generators/GenerationContext.java
```diff
@@ -78,29 +79,33 @@ static class LanguageGenerationConfiguration {
...
-        primitiveTypes);
+        primitiveTypes,
+        mappings);
   }
 
   GenerationContext(
       @Nonnull Set<LanguageGenerationConfiguration> languageConfs,
-      @Nonnull Map<String, String> primitiveTypes) {
+      @Nonnull Map<String, String> primitiveTypes,
+      @Nonnull Map<String, String> mappings) {
```
dslmeinte: Also here: please provide some documentation — maybe also for the class as a whole.

gradle-plugin/src/main/java/io/lionweb/gradleplugin/tasks/AbstractGenerationTask.java
```diff
@@ -38,7 +43,32 @@ public abstract class AbstractGenerationTask extends DefaultTask {
...
+    Set<File> classpath = getProject().getConfigurations().getByName("compileClasspath").resolve();
+    classpath.stream().filter(f -> f.getName().endsWith(".jar")).forEach(jar -> {
+      try(JarFile jarFile = new JarFile(jar)) {
+        Enumeration<JarEntry> entries = jarFile.entries();
+        while (entries.hasMoreElements()) {
+          JarEntry entry = entries.nextElement();
+          if (entry.getName().startsWith("META-INF/lionweb/")) {
+            InputStream inputStream = jarFile.getInputStream(entry);
+            JsonElement je = JsonParser.parseReader(new java.io.InputStreamReader(inputStream));
+            dependenciesChunks.add(new LowLevelJsonSerialization()
```
dslmeinte: An exception thrown for an individual file means loading all files fails, without any idea which file is the culprit: could you make that a bit more user-friendly?

PR #311, opened 2026-02-27
Performance improvements

ftomassetti:
These emerged from examining memory allocations in a complex project. These changes had a very significant impact, making the system way faster.

The main change consists in caching features: we check features in classifiers all the time, and calculating the features (including the inherited one) is happening very frequently.

Secondarily we avoid allocations and streams in `InMemoryServer.nodesByClassifier`, another frequently called method.

TODO:

- [x] Clear caches also when changing ancestors through dynamic methods

### DIFF
--- a/core/src/main/java/io/lionweb/language/Classifier.java
+++ b/core/src/main/java/io/lionweb/language/Classifier.java
@@ -118,52 +118,28 @@ public abstract class Classifier<T extends M3Node> extends LanguageEntity<T>
 
   public @Nonnull List<Property> allProperties() {
     if (cachedAllProperties == null) {
-      List<Property> result = new ArrayList<>();
-      for (Feature<?> f : allFeatures()) {
-        if (f instanceof Property) {
-          result.add((Property) f);
-        }
-      }
-      cachedAllProperties = Collections.unmodifiableList(result);
+      cachedAllProperties = filterFeatures(Property.class);
     }
     return cachedAllProperties;
   }
 
   public @Nonnull List<Containment> allContainments() {
     if (cachedAllContainments == null) {
-      List<Containment> result = new ArrayList<>();
-      for (Feature<?> f : allFeatures()) {
-        if (f instanceof Containment) {
-          result.add((Containment) f);
-        }
-      }
-      cachedAllContainments = Collections.unmodifiableList(result);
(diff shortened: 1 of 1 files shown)


client/src/main/java/io/lionweb/client/inmemory/InMemoryServer.java
```diff
@@ -193,13 +195,26 @@ public Map<ClassifierKey, ClassifierResult> nodesByClassifier(
...
-                  .map(n -> n.getID())
-                  .collect(Collectors.toSet()),
-              entry.getValue().size());
+      List<SerializedClassifierInstance> instances = entry.getValue();
+
+      // Calculate how many elements we will actually take
+      int actualLimit = (limit != null) ? limit : Integer.MAX_VALUE;
+      int targetSize = Math.min(instances.size(), actualLimit);
+
+      // Pre-allocate the HashSet considering Java's standard load factor (0.75)
```
dslmeinte: Isn’t this magic number a constant somewhere in the Java stdlib?

core/src/main/java/io/lionweb/language/Classifier.java
```diff
@@ -94,43 +103,69 @@ public Classifier(@Nullable Language language, @Nullable String name, @Nonnull S
...
   }
 
   public abstract @Nonnull List<Feature<?>> inheritedFeatures();
 
   public @Nonnull List<Property> allProperties() {
-    return allFeatures().stream()
-        .filter(f -> f instanceof Property)
-        .map(f -> (Property) f)
-        .collect(Collectors.toList());
+    if (cachedAllProperties == null) {
```
dslmeinte: It seems to me you should be able to DRY the code for the `all<X>` methods.

client/src/main/java/io/lionweb/client/inmemory/InMemoryServer.java
ftomassetti: Apparently yes, but it is private. I will define a constant

core/src/main/java/io/lionweb/language/Classifier.java
ftomassetti: Right, I did so for the 4 methods filtering features

PR #415, opened 2026-07-23
Fix serialization of references to LionCore-builtins for LionWeb 2024.1

### LINKED ISSUE #411: LionWeb 2024.1 references to builtins serialized wrong?
To my understanding, a property of string type should be serialized as

```json
{
  "references": [
    {
      "reference": {
        "language": "LionCore-M3",
        "version": "2024.1",
        "key": "Property-type"
      },
      "targets": [
        {
          "resolveInfo": "LionWeb.LionCore_builtins.String",
          "reference": null
        }
      ]
    }
  ],
}
```

but with lionweb-java v1.4.3, it's serialized as
```json
{
  "references": [
    {
      "reference": {
        "language": "LionCore-M3",
        "version": "2024.1",
        "key": "Property-type"
      },
      "targets": [
        {
          "resolveInfo": "LionWeb.LionCore_builtins.String",
          "reference": "LionCore-builtins-String-2024-1"
        }
      ]
    }
  ],
}
```

So `reference` should be `null`, but is set.
See https://github.com/LionWeb-io/specification/issues/279

ftomassetti:
Fix #411

### DIFF
--- a/core/src/main/java/io/lionweb/serialization/AbstractSerialization.java
+++ b/core/src/main/java/io/lionweb/serialization/AbstractSerialization.java
@@ -97,7 +97,7 @@ public abstract class AbstractSerialization {
 
   private final @Nonnull LionWebVersion lionWebVersion;
 
-  protected boolean builtinsReferenceDangling = false;
+  protected boolean builtinsReferenceDangling;
 
   protected AbstractSerialization() {
     this(LionWebVersion.currentVersion);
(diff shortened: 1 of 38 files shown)


dslmeinte (approved):
Looks good to me, assuming `ClassifierInstanceUtils` works as expected/assumed.

core/src/main/java/io/lionweb/serialization/AbstractSerialization.java
```diff
@@ -331,7 +336,13 @@ private void serializeReferences(
                           rv -> {
                             String referredID =
                                 rv.getReferred() == null ? null : rv.getReferred().getID();
+                            // References to M3 elements and built-ins are serialized without a
+                            // target reference id, relying on the resolveInfo to resolve them back
+                            // (specification section 3.4). We only drop the id when a resolveInfo
+                            // is
```
enikao: nitpicking: Is this formatting intended?

enikao (approved):

core/src/main/java/io/lionweb/serialization/AbstractSerialization.java
ftomassetti: spotless must have gone rogue... fixing

ftomassetti:
> Looks good to me, assuming `ClassifierInstanceUtils` works as expected/assumed.

Hopefully this is the case, and the test should (limitedly) verify that

ftomassetti:
Integration tests failing because of changes in them. Fixing them to make the build green before merging

PR #417, opened 2026-07-24
Fix annotation internal references

enikao:


enikao:
I found another bug similar to #417.

pjmolina (approved):
looks good!

ftomassetti:
@enikao I just released 1.4.5 with this fix, in case it is useful

